PR #5, opened 2026-04-03
Update README.md

simonepelosi:


simonepelosi:
This is temporary since Flatpack is not working atm

PR #43, opened 2026-05-13
feat: Tune Chrome audio flags for WSL2 and remove single-process mode

Wiibleyde:
This pull request refactors how Chromium launch arguments are constructed for the player, moving them into a dedicated helper function and updating the arguments to improve audio handling on Linux/WSL2. It also adds a unit test to ensure the correct flags are present.

**Refactoring and code organization:**

* Extracted the Chromium launch arguments from inline construction in `New` to a new helper function `launchArgs`, improving maintainability and readability (`internal/player/cdp/player.go`). [[1]](diffhunk://#diff-df65b0d60f24506fe46f9559365fe1e5da78dcf68cfb2826e098604b30e4c02cL99-R99) [[2]](diffhunk://#diff-df65b0d60f24506fe46f9559365fe1e5da78dcf68cfb2826e098604b30e4c02cR247-R289)

**Audio handling improvements for Linux/WSL2:**

* Added `--audio-buffer-size=4096` and `--disable-features=AudioServiceOutOfProcess` flags in `launchArgs` to address audio buffering and distortion issues on WSL2/PulseAudio setups (`internal/player/cdp/player.go`).

**Testing:**

* Introduced a new Linux-only test `TestLaunchArgs_WSL2AudioFlags` to verify that `launchArgs` includes the necessary audio flags and the correct Widevine path (`internal/player/cdp/player_test.go`).

PR #44, opened 2026-05-15
Add macOS support for Chrome/CDP playback

EzraCerpac:
## Summary

I added macOS support while keeping the existing Linux playback path intact.

This PR:
- splits the platform-specific startup code so Linux keeps the existing Chrome download, WebKit/GStreamer fallback, and MPRIS support
- adds a macOS CDP backend path that uses an installed Google Chrome app for MusicKit/Widevine playback
- adds macOS Chrome discovery with `VIBEZ_CHROME_PATH` / `CHROME_PATH` overrides
- replaces hardcoded `xdg-open` calls with platform-specific URL opening
- makes desktop icon/entry installation platform-specific
- adds macOS CI build/vet/test coverage and darwin release targets
- updates docs with minimal macOS setup notes

## Testing

I ran:

```sh
go test ./...
go vet ./...
CGO_ENABLED=0 go build ./...
go run . --demo
```

I also ran the macOS CDP launch probe with invalid MusicKit tokens:

```sh
VIBEZ_CDP_LAUNCH_PROBE=1 go test ./internal/player/cdp -run TestCDPLaunchProbeInvalidToken -v
```

Demo mode runs well on macOS.

I have not been able to test full Apple Music playback because I do not have an Apple Developer MusicKit key. The implementation reaches the Chrome/CDP setup path, but real DRM playback still needs verification by someone with valid Apple Music credentials and a developer token.

## Note

Most of these changes were made with AI assistance. I reviewed and validated the resulting diff locally, but I want to be transparent about how the patch was produced.


cmd/root_darwin.go
```diff
@@ -0,0 +1,149 @@
...
+
+func runPlatform(cfg *config.Config, iconPath string, opts tui.Options, onUserToken, onStorefront func(string)) error {
+	return runCDPFlow(cfg, iconPath, opts, onUserToken, onStorefront)
+}
+
+func runCDPFlow(cfg *config.Config, iconPath string, opts tui.Options, onUserToken, onStorefront func(string)) error {
+	prog := tea.NewProgram(tui.New(cfg, nil, nil, opts))
+	playerCh := make(chan *cdp.Player, 1)
+	runDone := make(chan struct{})
+	restartExe := make(chan string, 1)
```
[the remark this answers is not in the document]
EzraCerpac: I agree this flow has duplicated lifecycle logic. I’m leaving that refactor out of this PR for now because it would expand the macOS-support patch into a broader startup architecture change. The current split keeps Linux behavior easier to compare against the existing implementation while the macOS path is still being reviewed and tested. If you’d prefer, I can follow up in this PR and extract the shared CDP lifecycle behind small platform hooks.

simonepelosi:
Thank you so much for your first contribution ❤️ I will review it during this weekend! 

P.S. do we need chrome on MacOS ? I did not investigate too much on that yet

EzraCerpac:
> P.S. do we need chrome on MacOS ? I did not investigate too much on that yet

I looked into that. Music.app exposes AppleScript playback controls, but they are mostly remote-control style: play/pause/next/current track and library/playlist search. I don’t see an API to set Music.app’s queue from arbitrary Apple Music catalog IDs and mirror playback state the way vibez does through MusicKit JS. So I believe Music.app might work for a limited “control existing playback” backend, but not as a replacement for the CDP/MusicKit player in this PR. A native Swift MusicKit helper could be a future macOS backend, and this would be a nice solution, but it is a larger architecture change (and difficult to test for me without a developer key).

simonepelosi:
Yep totally agree, I was planning to do something in swift but I guess for now we can use Chrome. In future I would love to remove chrome everywhere if it's possibile, even on Linux. Btw, I have to find a way to share my developer key with contributors or a token generated using it.

simonepelosi (changes requested):
Good platform split — clean use of build tags, proper Chrome discovery on macOS, and the openurl abstraction is well-done. The duplication between darwin/linux runCDPFlow is the main concern. See inline comments.

cmd/root_darwin.go
```diff
@@ -0,0 +1,149 @@
...
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"syscall"
+	"time"
+
+	tea "charm.land/bubbletea/v2"
+	"github.com/simone-vibes/vibez/internal/auth"
```
simonepelosi: `runCDPFlow` in `root_darwin.go` and `root_linux.go` are ~90% identical (auth check, updater, EnsureBrowser, cdp.New, WaitReady, session-expired handler, lastfm). The only real difference is: Linux has WebKit fallback + MPRIS, and macOS skips those.

This will become a maintenance burden — any change to the CDP flow must be duplicated. Consider extracting shared CDP logic into a `cmd/cdp_common.go` (no build tag) that both platform files call, with platform-specific hooks injected via function params or an interface.

internal/player/cdp/browser.go
```diff
@@ -185,3 +185,45 @@ func runPlaywright() (*playwright.Playwright, error) {
...
+		// audio-only headless playback; Widevine CDM runs in a utility process
+		// and does not require GPU acceleration for audio DRM.
+		"--disable-gpu",
+		// Use /tmp for shared-memory segments instead of /dev/shm to avoid
+		// exhausting the (often small) tmpfs mounted there.
+		"--disable-dev-shm-usage",
+		// Cap the V8 JavaScript heap at 256 MB. MusicKit.js runs comfortably
+		// within this limit; without it Chrome can balloon to 500 MB+.
+		"--js-flags=--max-old-space-size=256",
+		// Disable background network activity (prefetch, DNS pre-resolve,
```
simonepelosi, line 229: `--ignore-certificate-errors` in the Linux Chrome args is a security concern. The comment mentions it's to work around `ERR_CERT_VERIFIER_CHANGED` from component updater swaps — but component update is already disabled via `--disable-component-update`. Is this flag still needed? If so, it means all HTTPS traffic from this Chrome instance (including Apple Music API calls) skips cert validation.

internal/player/cdp/browser_darwin.go
```diff
@@ -0,0 +1,119 @@
...
+		SkipInstallBrowsers: true,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("playwright driver: %w", err)
+	}
+	return pw, nil
+}
+
+func chromeLaunchArgs() []string {
+	return []string{
```
simonepelosi, line 111: The macOS `chromeLaunchArgs()` doesn't include `--widevine-path=...`. On Linux, Widevine is explicitly pointed at the downloaded Chrome's bundled CDM. On macOS, you rely on the system Chrome's built-in Widevine — is this guaranteed to work with Playwright's CDP connection? Have you verified DRM playback actually works, or is this the "real DRM playback still needs verification" gap mentioned in the PR description?

internal/player/cdp/browser_darwin.go
```diff
@@ -0,0 +1,119 @@
...
+	"path/filepath"
+
+	playwright "github.com/playwright-community/playwright-go"
+)
+
+const chromeInstallHelp = "install Google Chrome from https://www.google.com/chrome/ or set VIBEZ_CHROME_PATH/CHROME_PATH"
+
+func baseDir() string {
+	if d := os.Getenv("XDG_CACHE_HOME"); d != "" {
+		return filepath.Join(d, "vibez")
```
simonepelosi, line 17: `baseDir()` checks `XDG_CACHE_HOME` on macOS. While technically harmless, this is a Linux convention — macOS uses `~/Library/Caches`. If a user has `XDG_CACHE_HOME` set (e.g., from cross-platform dotfiles), vibez would use a non-standard location on macOS. Consider removing the XDG check on darwin and always using `~/Library/Caches/vibez`.

.github/workflows/ci.yml
```diff
@@ -60,3 +60,23 @@ jobs:
       - name: Coverage summary
         run: |
           go tool cover -func=coverage.out | tail -1 | awk '{print "Total coverage: " $3}' >> $GITHUB_STEP_SUMMARY
+
+  test-macos:
+    name: build-vet-test-macos
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v4
```
simonepelosi, line 68: The macOS CI job uses `CGO_ENABLED=0 go build ./...` but then runs `go test ./...` without CGO_ENABLED=0. If some tests depend on CGO-enabled packages (unlikely for macOS but worth checking), this could silently skip tests. Also, building with CGO_ENABLED=0 means the CI build artifact may differ from what users actually run. Is this intentional (e.g., to avoid needing Xcode tools on CI)?

EzraCerpac:
Thanks for the review! I agree with the main points, especially the duplicated startup flow. I’ll update the PR to extract this into common code, remove the Linux `--ignore-certificate-errors`, update the macOS cache path, and tighten the macOS CI CGO coverage.

On Widevine: yes, that is the remaining real-playback verification gap from the PR description. The macOS path relies on installed Google Chrome’s bundled Widevine. I’ve verified Chrome/CDP launch and the pre-auth MusicKit wiring locally, but I still can’t honestly claim full DRM playback works without a MusicKit developer token.

.github/workflows/ci.yml
EzraCerpac: The `CGO_ENABLED=0` build was intentional because the darwin GoReleaser build ships with `CGO_ENABLED=0`. You’re right that the CI should make this clearer and cover that mode consistently. I’ll add a `CGO_ENABLED=0 go test ./...` step while keeping the normal macOS test too.

internal/player/cdp/browser_darwin.go
EzraCerpac: Good point. I’ll switch cache discovery to `os.UserCacheDir()` and append `vibez`. That keeps Linux behavior equivalent to the existing `XDG_CACHE_HOME` / `~/.cache` logic, and gives native `$HOME/Library/Caches` on macOS.

EzraCerpac:
I pushed an update with the cleanups and rebased the PR onto current `main`.

Main changes:
- extracted the shared CDP startup flow
- preserved the WSL2 audio changes from #43
- removed the broad Linux cert-ignore flag
- switched cache paths to Go’s `os.UserCacheDir()`
- tightened the README wording and macOS CI/test coverage

simonepelosi:
That looks great now! Thank you so much, let me test it on my Mac with the Dev key to confirm and we are good to go! 

Also nice: removed `--single-process` from Linux args (was a stability risk), added WSL audio buffering support, and the `cdpPlatformHooks` pattern is extensible for future platforms.

simonepelosi:
It works! ❤️

EzraCerpac:
Amazing, thanks for merging this! I’m really happy to see macOS support land and once a macOS release is published, I’m looking forward to using vibez myself!

PR #45, opened 2026-05-15
Fix Apple Music playlist browsing

elijah-rou:
## Summary
- include Favorites in library playlists
- add back navigation from playlist track view
- load playlist tracks reliably with encoded IDs and visible errors
- remove obsolete provider playlist regression tests that asserted the old playlist loading path; behavior is now covered at the library/TUI boundary

## Tests
- `PKG_CONFIG_PATH=$PWD/pkg-config go test ./internal/tui ./internal/tui/views ./internal/provider/apple`
- `node internal/player/web/musickit_test.mjs`


simonepelosi (changes requested):
Good PR overall — clean URL escaping, solid race condition fix, good test coverage. A few concerns need addressing before merge. See inline comments.

internal/tui/views/library.go
```diff
@@ -87,6 +88,14 @@ func (m *LibraryModel) Height() int     { return m.height }
 func (m *LibraryModel) DrillErr() error { return m.drillErr }
 func (m *LibraryModel) LoadErr() error  { return m.loadErr }
 
+func (m *LibraryModel) Back() bool {
+	if m.pane != paneTracks {
+		return false
+	}
+	m.pane = paneList
```
simonepelosi: `Back()` sets `m.pane = paneList` but does not reset `m.drillPlaylist`. Since the new stale-response check on line ~141 compares `msg.playlist.ID != m.drillPlaylist.ID`, if a user goes back and then re-drills into the *same* playlist, a late response from the previous fetch would pass the guard and be accepted.

Consider adding `m.drillPlaylist = provider.Playlist{}` here (or resetting it when the new drill starts).

internal/provider/apple/apple.go
```diff
@@ -602,6 +621,89 @@ func (a *AppleProvider) GetCatalogPlaylistTracks(ctx context.Context, playlistID
...
+func (a *AppleProvider) getSongRatings(ctx context.Context, ids []string) (map[string]bool, error) {
+	loved := make(map[string]bool, len(ids))
+	for start := 0; start < len(ids); start += ratingBatchSize {
+		end := min(start+ratingBatchSize, len(ids))
+		ep := "/me/ratings/songs?ids=" + url.QueryEscape(strings.Join(ids[start:end], ","))
+		req, err := a.newRequest(ctx, http.MethodGet, ep)
+		if err != nil {
+			return nil, err
+		}
+		resp, err := a.client.Do(req) //nolint:gosec // G704: URL is constructed from config, not user input
```
simonepelosi: This `"r."` prefix stripping maps rating resource IDs back to catalog song IDs — but this behavior is not documented in Apple's public API docs. Could you add a brief comment explaining this mapping (e.g., empirically observed, or link to source)? If Apple ever changes this prefix the feature would silently break with no errors.

internal/provider/apple/apple.go
```diff
@@ -602,6 +621,89 @@ func (a *AppleProvider) GetCatalogPlaylistTracks(ctx context.Context, playlistID
...
+	seen := make(map[string]bool, len(tracks))
+	for _, track := range tracks {
+		catalogID := track.CatalogID
+		if catalogID == "" && !strings.HasPrefix(track.ID, "i.") {
+			catalogID = track.ID
+		}
+		if catalogID == "" {
+			continue
+		}
+		trackCatalogID[track.ID] = catalogID
```
simonepelosi, line 645: This fetches ALL library songs sequentially, then batch-queries ratings in chunks of 100. For a large library (10k+ songs) this is ~100+ sequential API calls under a 20s timeout inherited from the caller — likely to timeout.

A few options: (a) bounded-parallel rating batches (e.g., 5 concurrent), (b) increase timeout for this specific code path, or (c) at minimum add a comment documenting the limitation. What's the expected library size for target users?

internal/provider/apple/apple.go
```diff
@@ -510,12 +513,28 @@ func (a *AppleProvider) GetLibraryPlaylists(ctx context.Context) ([]provider.Pla
...
 
+func hasFavoritesPlaylist(playlists []provider.Playlist) bool {
+	for _, pl := range playlists {
+		name := strings.ToLower(strings.TrimSpace(pl.Name))
+		if name == "favorites" || name == "favorite songs" {
+			return true
+		}
+	}
+	return false
+}
```
simonepelosi, line 535: This only matches English names ("favorites", "favorite songs"). Apple Music localizes playlist names — e.g. "Favoritos" (Spanish), "Favoris" (French). Is this intentional for v1? If so, a brief `// TODO: handle localized names` comment would clarify the scope.

internal/provider/apple/apple_test.go
```diff
@@ -923,3 +923,54 @@ func TestGetSongRating_InvalidJSON_ReturnsFalse(t *testing.T) {
 		t.Error("GetSongRating = true for invalid JSON, want false")
 	}
 }
```
simonepelosi, line 927: Commit `19d1b49` removes playlist regression tests but the PR description doesn't explain why. Were these testing behavior that changed with this PR? Could you add a note justifying the removal?

elijah-rou:
Pushed the playlist follow-up changes: stale playlist responses are request-gated, Favorites rating lookups use bounded concurrency, the Apple `r.` prefix behavior is documented, the Favorites-name limitation is called out, and playlist coverage is back in provider/TUI tests.

elijah-rou:
Rebased `fix/playlists` onto latest upstream `main`, kept the playlist fixes intact, updated the test-removal note, and reran the focused/full Go test suites.

internal/tui/views/library.go
elijah-rou: I now clear the drill state on back navigation and gate playlist responses by request ID, so stale async results are ignored.

internal/provider/apple/apple.go
elijah-rou: Added a note around the observed Apple `r.` ratings prefix and kept the stripping limited to that exact prefix.

internal/provider/apple/apple.go
elijah-rou: Changed this to bounded parallel rating lookups, currently 5 workers, so large libraries do not serialize every ratings call.

internal/provider/apple/apple.go
elijah-rou: Documented this as a v1 limitation. Apple does not expose a stable built-in Favorites playlist identifier in these responses, so suppression is still based on known English names.

internal/provider/apple/apple_test.go
elijah-rou: Added coverage back around the playlist behavior in the Apple provider and TUI tests.

simonepelosi (dismissed):
All good — thanks for the thorough fixes! The bounded-parallel ratings fetching, drillPlaylist reset in Back(), and the explanatory comments all look solid. Appreciate the drillRequestID generation guard too. 🎵

simonepelosi:
@elijah-rou Can you fix the conflicts ? <3

elijah-rou:
> @elijah-rou Can you fix the conflicts ? <3

Yes ma'am! Will do it today

elijah-rou:
@simonepelosi I found a bug in demo mode I need to fix, hold off merge for now!

simonepelosi:
@elijah-rou Oh great! Thanks, I see the CI is failing

elijah-rou:
@simonepelosi okay! should be good to go! may wanna do a quick relook

simonepelosi:
Thank you so much again ❤️ It looks great!

PR #46, opened 2026-05-15
Add configurable Apple Music audio quality

elijah-rou:
## Summary
- add `audio_quality` config with high/standard options
- add `:quality` command for 256/64 kbps AAC
- clearly reject lossless/unsupported values for MusicKit web backend

## Tests
- `PKG_CONFIG_PATH=$PWD/pkg-config go test ./...`
- `PKG_CONFIG_PATH=$PWD/pkg-config go build ./...`
- `node internal/player/web/musickit_test.mjs`

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -55,6 +55,11 @@ func runTUI(_ *cobra.Command, _ []string) error {
 		return fmt.Errorf("loading config: %w", err)
 	}
 
+	audioBitrateKbps, err := cfg.AudioBitrateKbps()
+	if err != nil {
+		return err
+	}
+
 	opts := tui.Options{MemProfiling: memProfiling}
 
 	// Apply theme before creating any TUI model so all panels pick up the palette.
@@ -104,5 +109,6 @@ func runTUI(_ *cobra.Command, _ []string) error {
 		}
 	}
 
-	return runPlatform(cfg, iconPath, opts, onUserToken, onStorefront)
+	return runPlatform(cfg, iconPath, opts, onUserToken, onStorefront, audioBitrateKbps)
+
 }
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -93,6 +93,7 @@ func runTUI(_ *cobra.Command, _ []string) error {
 	// but lets the DE resolve the icon via the MPRIS DesktopEntry property.
 	iconPath := assets.InstallIcon()
 	assets.InstallDesktopEntry()
+	opts.IconPath = iconPath
 
 	onUserToken := func(token string) {
 		cfg.AppleUserToken = token
--- a/internal/player/webkit/player.go
+++ b/internal/player/webkit/player.go
@@ -231,7 +231,7 @@ func (p *Player) SetAudioBitrate(kbps int) error {
 	if err := audioquality.Validate(kbps); err != nil {
 		return err
 	}
-	return fmt.Errorf("WebKit/GStreamer uses fixed 30 s preview URLs and cannot change bitrate; MusicKit JS/web playback max is 256 kbps AAC")
+	return player.ErrAudioBitrateRequiresRestart
 }
 
 func (p *Player) SetQueue(ids []string) error {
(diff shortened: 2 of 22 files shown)


simonepelosi:
Thank you so much for your first contribution ❤️ I will review it during this weekend

elijah-rou:
Of course! Recently installed Omarchy and it's a great fir there, so I thought I'd shoot through some things as I use it! Great work btw!

simonepelosi (changes requested):
Clean, well-tested feature addition. Good input validation and clear error messages for unsupported values. A few concerns inline.

internal/player/webkit/player.go
```diff
@@ -224,6 +225,13 @@ func (p *Player) SetVolume(v float64) error {
 	return nil
 }
 
+func (p *Player) SetAudioBitrate(kbps int) error {
+	if err := audioquality.Validate(kbps); err != nil {
+		return err
+	}
+	return fmt.Errorf("WebKit/GStreamer uses fixed 30 s preview URLs and cannot change bitrate; MusicKit JS/web playback max is 256 kbps AAC")
```
simonepelosi: This validates the input and then unconditionally returns an error. From the user's perspective, `:quality high` (a valid value) returns an error on the WebKit backend, which is confusing.

Consider: (a) storing the preference in config even if it can't be applied at runtime (so next CDP launch uses it), or (b) returning a more clearly worded "saved for next session but cannot apply on WebKit backend" message instead of a hard error.

internal/tui/model.go
```diff
@@ -1195,6 +1210,31 @@ func (m *Model) executeCommand(cmd string) tea.Cmd {
...
+			m.errExpiry = time.Now().Add(4 * time.Second)
+			return nil
+		}
+		m.appendLog(fmt.Sprintf("[quality] → %d kbps AAC", kbps))
+		return func() tea.Msg {
+			if m.player == nil {
+				return errMsg{fmt.Errorf("no player")}
+			}
+			if err := m.player.SetAudioBitrate(kbps); err != nil {
+				return errMsg{err}
```
simonepelosi, line 1247: `m.player` is accessed inside the returned `func() tea.Msg` closure. Is `m.player` guaranteed non-nil at execution time (not just at closure creation time)? If it can be set to nil (e.g., during teardown), this could panic. The nil check inside helps, but there's a TOCTOU gap between the check and `m.player.SetAudioBitrate(kbps)`.

cmd/root.go
```diff
@@ -96,7 +101,7 @@ func runTUI(_ *cobra.Command, _ []string) error {
 	// Install icon and .desktop entry on every launch (idempotent, best-effort).
 	// The .desktop file has NoDisplay=true so it stays invisible to app launchers
 	// but lets the DE resolve the icon via the MPRIS DesktopEntry property.
-	iconPath := assets.InstallIcon()
+	assets.InstallIcon()
```
simonepelosi: `iconPath` was removed from the function signature and `assets.InstallIcon()` return value is now discarded. Is `iconPath` still used elsewhere (e.g., passed to MPRIS)? If it was only used here, fine — but verify no downstream consumer needs it.

elijah-rou:
Rebased this branch onto latest upstream `main`, kept the audio-quality fixes across the new platform split, and reran focused tests plus `make test`.

internal/player/webkit/player.go
elijah-rou: WebKit now treats quality changes as saved preference only. The command persists valid values, while the current backend reports that it cannot switch bitrate live.

internal/tui/model.go
elijah-rou: The command now captures the player before creating the Bubble Tea command, so it does not re-read `m.player` after teardown.

cmd/root.go
elijah-rou: Restored the installed icon path flow. `InstallIcon()` now feeds `opts.IconPath`, which is passed through startup for notification/MPRIS consumers.

simonepelosi (dismissed):
Looks great, thanks for addressing the feedback! The ErrAudioBitrateSavedPreferenceOnly pattern is much clearer for WebKit users, and the playerCmd refactor to pass the player by argument is a nice safety improvement across the board. Ship it 🚀

PR #47, opened 2026-05-15
Add library sections for songs albums artists playlists

elijah-rou:
## Summary
- show Songs, Albums, Artists, Playlists when opening Library
- add drill-down/playback for songs, album groups, artist groups, playlists
- preserve playlist behavior with back navigation

## Tests
- `go test ./internal/tui/views ./internal/tui`
- `go test ./...`

### DIFF
--- a/internal/tui/views/library.go
+++ b/internal/tui/views/library.go
@@ -13,12 +13,22 @@ import (
 	"github.com/simone-vibes/vibez/internal/tui/styles"
 )
 
-// libraryPane controls which level of the drill-down is visible.
 type libraryPane int
 
 const (
-	paneList   libraryPane = iota // top-level tab list
-	paneTracks                    // tracks inside a selected playlist
+	paneSections libraryPane = iota
+	paneItems
+	paneTracks
+	paneList = paneItems
+)
+
+type librarySection int
+
+const (
+	sectionSongs librarySection = iota
+	sectionAlbums
+	sectionArtists
+	sectionPlaylists
 )
 
 type libraryLoadedMsg struct {
@@ -26,10 +36,21 @@ type libraryLoadedMsg struct {
 	err       error
 }
 
+type libraryTracksLoadedMsg struct {
+	tracks []provider.Track
+	err    error
+}
+
 type playlistTracksMsg struct {
 	playlist provider.Playlist
 	tracks   []provider.Track
 	err      error
+}
+
+type trackGroup struct {
+	title  string
+	desc   string
+	tracks []provider.Track
 }
 
 type LibraryModel struct {
@@ -39,15 +60,22 @@ type LibraryModel struct {
 	list     list.Model
 	spinner  spinner.Model
 
-	playlists []provider.Playlist
+	pane            libraryPane
+	selectedSection librarySection
+	libraryTracks   []provider.Track
+	tracksLoaded    bool
+	playlists       []provider.Playlist
+	playlistsLoaded bool
+	albums          []trackGroup
+	artists         []trackGroup
 
-	// Drill-down into a playlist's tracks.
-	pane          libraryPane
-	drillPlaylist provider.Playlist
-	drillTracks   []provider.Track
-	drillLoading  bool
-	drillErr      error
-	drillList     list.Model
+	drillTitle     string
+	drillPlaylist  provider.Playlist
+	drillTracks    []provider.Track
+	drillLoading   bool
+	drillErr       error
+	drillList      list.Model
+	tracksBackPane libraryPane
 
 	width  int
 	height int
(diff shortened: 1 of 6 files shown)


simonepelosi (changes requested):
Solid feature — clean state machine, good separation of sections/items/tracks. A few concerns need addressing, mostly around the confusing paneList alias and missing context timeouts.

internal/tui/views/library.go
```diff
@@ -13,41 +13,69 @@ import (
...
 
 const (
-	paneList   libraryPane = iota // top-level tab list
-	paneTracks                    // tracks inside a selected playlist
+	paneSections libraryPane = iota
+	paneItems
+	paneTracks
+	paneList = paneItems
+)
+
```
simonepelosi, line 29: `paneList = paneItems` is an alias that creates confusion. In `handleKey` under `paneTracks`, there's `m.pane = paneList` which looks like it's going back to some "list" pane but is actually `paneItems`. 

Meanwhile PR #45 references `paneList` as the back target for `Back()`. If both PRs merge, the semantics diverge. Suggest: remove the alias and use `paneItems` consistently, or rename to something clearer.

internal/tui/views/library.go
```diff
@@ -95,6 +113,15 @@ func (m *LibraryModel) SetSize(w, h int) {
 	m.drillList.SetSize(w, listH)
 }
 
+func (m *LibraryModel) loadLibraryTracks() tea.Cmd {
+	m.loading = true
+	prov := m.provider
```
simonepelosi, line 159: This uses `context.Background()` with no timeout. For large libraries this could hang indefinitely. PR #45 adds 20s timeouts to the playlist load paths — should be consistent here too.

internal/tui/views/library.go
```diff
@@ -115,105 +142,206 @@ func (m *LibraryModel) loadPlaylistTracks(pl provider.Playlist) tea.Cmd {
...
+	first := allTracks[0]
+	tracks := append([]provider.Track{}, allTracks...)
+	return func() tea.Msg { return PlayTracksMsg{IDs: ids, Tracks: tracks, Track: &first} }
+}
+
+func (m *LibraryModel) showSections() {
+	items := []list.Item{sectionItem{sectionSongs}, sectionItem{sectionAlbums}, sectionItem{sectionArtists}, sectionItem{sectionPlaylists}}
+	m.list.SetItems(items)
+	m.list.Select(0)
+}
```
simonepelosi, line 403: `panic("unknown library section")` in multiple places (`sectionTitle`, `openSelectedSection`, `routeLoadedLibraryTracks`). If a new section is ever added and one switch is missed, this crashes the whole TUI. Consider returning an error or a safe default instead of panicking in a UI layer.

internal/tui/views/library.go
```diff
@@ -115,105 +142,206 @@ func (m *LibraryModel) loadPlaylistTracks(pl provider.Playlist) tea.Cmd {
...
+			m.playlistsLoaded = true
+			m.showPlaylists()
 		}
 		return m, nil
-
 	case playlistTracksMsg:
 		m.drillLoading = false
 		m.drillErr = msg.err
+		m.drillPlaylist = msg.playlist
 		if msg.err == nil {
```
simonepelosi, line 233: Once `tracksLoaded = true`, the library tracks are cached forever — `openSelectedSection` skips re-fetching. If a user adds songs to their library during a session, they won't appear until restart. Is this intentional? If so, a comment would help. If not, consider a TTL or a manual refresh mechanism.

elijah-rou:
Rebased this branch onto latest upstream `main`, kept the library-section fixes intact, and reran focused TUI tests plus `make test`.

internal/tui/views/library.go
elijah-rou: Removed the alias and now use `paneItems` directly for the list pane/back target.

internal/tui/views/library.go
elijah-rou: Switched this load to a bounded timeout context, matching the playlist path.

internal/tui/views/library.go
elijah-rou: Removed the UI panic and route unknown sections through explicit error handling instead.

internal/tui/views/library.go
elijah-rou: Added an explicit refresh path so section results can be reloaded during the session instead of requiring restart.

simonepelosi (approved):
Thanks for addressing all the feedback! The paneList alias cleanup, context timeouts, panic removal, and the 5-minute TTL with generation-based stale response guarding all look solid. Nice work 👍

PR #48, opened 2026-05-22
Add ASCII album art to now-playing and demo artwork

### LINKED ISSUE #33: feat: ASCII album art in the now-playing panel
## Summary

Render the current track's artwork as Unicode block art directly inside the now-playing section of the TUI.

## Behaviour

- Download the artwork URL (already in `player.State.Track.ArtworkURL`) to an in-memory buffer
- Convert the image to a grid of coloured Unicode half-block characters (▀ / ▄) using 24-bit ANSI colour codes
- Render it on the left side of the now-playing panel, with title/artist/progress on the right
- Size adapts to the available terminal width/height
- Falls back gracefully (no art shown) if the terminal doesn't support true colour or the download fails

## Implementation

- `internal/tui/art/art.go` — pure Go image → block-art converter (no extra deps beyond `image/jpeg`, `image/png` from stdlib)  
- Cache the last rendered artwork by URL so resizes don't re-download

## Why it's cool

Most TUI music players show a plain text track name. Block art makes the now-playing panel look genuinely great, especially with a themed colour scheme.

elijah-rou:
Summary:
- Adds ASCII/terminal album art rendering to the now-playing UI with bounded sizing and fallback behavior.
- Adds demo track artwork so the demo mode can exercise the now-playing artwork path.
- Covers artwork rendering, fallback, race/bounds behavior, and demo/provider/TUI integration.

Tests run:
- `go test ./...`

Notes:
- No bundled commercial album art. Demo tracks use generated placeholder artwork URLs.

Closes #33


elijah-rou:
will open another when ready, opened early by mistake

PR #62, opened 2026-06-07
feat: local tracks able to play music from your filesystem

### LINKED ISSUE #61: feat: local tracks provider (M4A, FLAC, MP3, OGG)
Hello i have been using this for a while now (its been 2.5 weeks lol) and now I would like to work on the local tracks feature from the roadmap.

The idea is to add a provider/local package that scans a configured 
music directory and surfaces tracks in the library panel, with playback 
through the existing gst/ layer using "file://" URIs. No DRM or auth needed.

Rough plan:
- internal/provider/local/ - scan dir, read metadata, implement provider interface
- internal/player/local/ - wrap gst/ to play file:// URIs
- internal/config/ - add a MusicDir field
- cmd/ - wire a --local flag similar to how --demo is wired today

Before I start writing code I wanted to check:
- Does this approach align with what you had in mind?
- Any preference on the metadata library for reading ID3/FLAC tags?
- Should the directory scan be recursive?

Happy to discuss before opening a PR.

GurKalra:
Closes #61 

### What
Adds a local music provider and player so vibez can be used without 
Apple credentials by playing audio files from the filesystem.

### Why
Requested in the roadmap. Useful for users who want a terminal music 
player for their local library without needing an Apple Music 
subscription.

### Changes
- `internal/provider/local/` - scans a directory recursively for MP3,
  FLAC, M4A and OGG files, reads metadata via `dhowden/tag`, implements
  the full provider interface including genre-based recommendations
- `internal/player/local/` - GStreamer-backed player for `file://` URIs
  with position/duration polling every 500ms
- `internal/config/` - adds `MusicDir` field
- `internal/tui/` - adds `InitialTracks` to `Options` to pre-populate
  the queue on launch
- `cmd/root.go` - wires `--local` and `--music-dir` flags, prompts for
  directory on first run and saves to config
- `Makefile` - adds `check-deps` target that detects missing system
  libraries before building with helpful error messages

### How to test
```bash
make build
./vibez --local --music-dir /path/to/your/music
```
First run with no `--music-dir` prompts for the directory and saves it.
Subsequent runs go straight to playing.

#### Notes
- Equalizer is stubbed. It needs 10-band EQ added to `gst/` pipeline first
- `golangci-lint` could not run locally due to `go 1.26` toolchain not
  yet available. Will be verified by CI.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1,10 +1,12 @@
 package cmd
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	tea "charm.land/bubbletea/v2"
 	"github.com/simone-vibes/vibez/internal/assets"
@@ -89,8 +91,8 @@ func runTUI(_ *cobra.Command, _ []string) error {
 		// before launching the TUI.
 		if cfg.MusicDir == "" {
 			fmt.Print("Enter path to your music directory: ")
-			var dir string
-			fmt.Scanln(&dir)
+			dir, _ := bufio.NewReader(os.Stdin).ReadString('\n')
+			dir = strings.TrimSpace(dir)
 			if dir == "" {
 				return fmt.Errorf("no music directory provided")
 			}
@@ -105,7 +107,7 @@ func runTUI(_ *cobra.Command, _ []string) error {
 		assets.InstallDesktopEntry()
 		prov, err := localProvider.New(cfg.MusicDir)
 		if err != nil {
-			return fmt.Errorf("local provider: %w",err) 
+			return fmt.Errorf("local provider: %w", err)
 		}
 		plyr, err := localPlayer.New()
 		if err != nil {
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -11,7 +12,9 @@ import (
 	"github.com/simone-vibes/vibez/internal/config"
 	"github.com/simone-vibes/vibez/internal/lastfm"
 	demoPlayer "github.com/simone-vibes/vibez/internal/player/demo"
+	localPlayer "github.com/simone-vibes/vibez/internal/player/local"
 	demoProvider "github.com/simone-vibes/vibez/internal/provider/demo"
+	localProvider "github.com/simone-vibes/vibez/internal/provider/local"
 	"github.com/simone-vibes/vibez/internal/tui"
 	"github.com/simone-vibes/vibez/internal/tui/styles"
 	"github.com/spf13/cobra"
@@ -22,6 +25,8 @@ var debug bool
 var memProfiling bool
 var demo bool
 var noUpdate bool
+var local bool
+var musicDir string
 
 var rootCmd = &cobra.Command{
 	Use:   "vibez",
@@ -43,6 +48,8 @@ func init() {
 	rootCmd.PersistentFlags().BoolVar(&memProfiling, "mem-profiling", false, "show live RSS for vibez and its Chrome helper in the header")
 	rootCmd.PersistentFlags().BoolVar(&demo, "demo", false, "run with built-in fake data — no Apple account or internet required")
 	rootCmd.PersistentFlags().BoolVar(&noUpdate, "no-update", false, "skip automatic update check on startup")
+	rootCmd.PersistentFlags().BoolVar(&local, "local", false, "run with local music files (no Apple account required)")
+	rootCmd.PersistentFlags().StringVar(&musicDir, "music-dir", "", "path to your music directory (saved to config)")
 }
 
 func runTUI(_ *cobra.Command, _ []string) error {
@@ -68,6 +75,62 @@ func runTUI(_ *cobra.Command, _ []string) error {
 		fmt.Fprintf(os.Stderr, "debug: theme: %v (falling back to default)\n", themeErr)
 	}
 	styles.Apply(theme)
+
+	if local {
+		// if --music-dir was passed, it will save to config.
+		if musicDir != "" {
+			cfg.MusicDir = musicDir
+			if saveErr := cfg.Save(cfgFile); saveErr != nil && debug {
+				fmt.Fprintf(os.Stderr, "debug: saving music dir: %v\n", saveErr)
+			}
+		}
+
+		// If musicDir is still empty, prompt the user on the command line
+		// before launching the TUI.
+		if cfg.MusicDir == "" {
+			fmt.Print("Enter path to your music directory: ")
+			var dir string
+			fmt.Scanln(&dir)
+			if dir == "" {
+				return fmt.Errorf("no music directory provided")
+			}
+			cfg.MusicDir = dir
+			if saveErr := cfg.Save(cfgFile); saveErr != nil && debug {
+				fmt.Fprintf(os.Stderr, "debug: saving music dir: %v\n", saveErr)
+			}
+			fmt.Println("Music directory saved. You can change it anytime in Settings.")
+		}
+
+		iconPath := assets.InstallIcon()
+		assets.InstallDesktopEntry()
+		prov, err := localProvider.New(cfg.MusicDir)
+		if err != nil {
+			return fmt.Errorf("local provider: %w",err) 
+		}
+		plyr, err := localPlayer.New()
+		if err != nil {
+			return fmt.Errorf("local player: %w", err)
+		}
+		tracks, err := prov.GetLibraryTracks(context.Background())
+		if err != nil {
+			return fmt.Errorf("loading local tracks: %w", err)
+		}
+		plyr.LoadTracks(tracks)
+		// Auto-play from the first track on launch.
+		if len(tracks) > 0 {
+			ids := make([]string, len(tracks))
+			for i, t := range tracks {
+				ids[i] = t.ID
+			}
+			_ = plyr.SetQueue(ids)
+		}
+		opts.IconPath = iconPath
+		opts.InitialTracks = tracks
+		opts.Backend = "Local mode · playing from " + cfg.MusicDir
+		prog := tea.NewProgram(tui.New(cfg, prov, plyr, opts))
+		_, err = prog.Run()
+		return err
+	}
 
 	if demo {
 		iconPath := assets.InstallIcon()
--- a/internal/player/local/player.go
+++ b/internal/player/local/player.go
@@ -17,12 +17,13 @@ import (
 
 // Player implements player.Player for local audio files using GStreamer.
 type Player struct {
-	gst *gst.Player
-	mu sync.RWMutex
+	gst   *gst.Player
+	mu    sync.RWMutex
 	state player.State
-	subs []chan player.State
+	subs  []chan player.State
 	queue []provider.Track
-	idx int
+	idx   int
+	done  chan struct{}
 }
 
 // New creates a local Player backed by a GStreamer pipeline.
@@ -31,14 +32,14 @@ func New() (*Player, error) {
 	if err != nil {
 		return nil, fmt.Errorf("local player: %w", err)
 	}
-	p := &Player{gst: gstPlayer}
-	p.gst.OnEOS(func () {
+	p := &Player{gst: gstPlayer, done: make(chan struct{})}
+	p.gst.OnEOS(func() {
 		_ = p.Next()
 	})
 	p.gst.OnError(func(e error) {
 		p.mu.Lock()
+		p.state.Error = e.Error()
 		s := p.state
-		s.Error = e.Error()
 		p.mu.Unlock()
 		p.broadcast(s)
 	})
@@ -50,7 +51,7 @@ func (p *Player) broadcast(s player.State) {
 	p.mu.RLock()
 	defer p.mu.RUnlock()
 	for _, ch := range p.subs {
-		select  {
+		select {
 		case ch <- s:
 		default:
 		}
@@ -169,11 +170,17 @@ func (p *Player) SetQueue(ids []string) error {
 	}
 	// Finding the first ID in the full queue and starting from there.
 	p.mu.Lock()
+	found := false
 	for i, t := range p.queue {
 		if t.ID == ids[0] {
 			p.idx = i
+			found = true
 			break
 		}
+	}
+	if !found || len(p.queue) == 0 {
+		p.mu.Unlock()
+		return nil
 	}
 	t := p.queue[p.idx]
 	p.mu.Unlock()
@@ -233,9 +240,12 @@ func (p *Player) RemoveFromQueue(idx int) error {
 
 func (p *Player) MoveInQueue(from, to int) error {
 	p.mu.Lock()
-	if from >= 0 && from < len(p.queue) && to >=0 && to < len(p.queue) {
+	if from >= 0 && from < len(p.queue) && to >= 0 && to < len(p.queue) {
 		t := p.queue[from]
 		p.queue = append(p.queue[:from], p.queue[from+1:]...)
+		if from < to {
+			to--
+		}
 		p.queue = append(p.queue[:to], append([]provider.Track{t}, p.queue[to:]...)...)
 	}
 	p.mu.Unlock()
@@ -257,10 +267,10 @@ func (p *Player) ClearQueue() error {
 
 func (p *Player) GetState() (*player.State, error) {
 	s := p.currentState()
-	return &s,nil
+	return &s, nil
 }
 
-func (p *Player) Subscribe() <- chan player.State {
+func (p *Player) Subscribe() <-chan player.State {
 	ch := make(chan player.State, 8)
 	p.mu.Lock()
 	p.subs = append(p.subs, ch)
@@ -269,6 +279,7 @@ func (p *Player) Subscribe() <- chan player.State {
 }
 
 func (p *Player) Close() error {
+	close(p.done)
 	p.gst.Destroy()
 	return nil
 }
@@ -299,18 +310,22 @@ func tracksForIDs(tracks []provider.Track, ids []string) []provider.Track {
 func (p *Player) pollState() {
 	ticker := time.NewTicker(500 * time.Millisecond)
 	defer ticker.Stop()
-	for range ticker.C {
-		p.mu.RLock()
-		playing := p.state.Playing
-		p.mu.RUnlock()
-		if !playing {
-			continue
+	for {
+		select {
+		case <-ticker.C:
+			p.mu.RLock()
+			playing := p.state.Playing
+			p.mu.RUnlock()
+			if !playing {
+				continue
+			}
+			p.mu.Lock()
+			p.state.Position = p.gst.Position()
+			s := p.state
+			p.mu.Unlock()
+			p.broadcast(s)
+		case <-p.done:
+			return
 		}
-		p.mu.Lock()
-		pos := p.gst.Position()
-		p.state.Position = pos
-		s := p.state
-		p.mu.Unlock()
-		p.broadcast(s)
 	}
 }
--- a/internal/player/local/player.go
+++ b/internal/player/local/player.go
@@ -1,0 +1,316 @@
+//go:build linux
+
+// Package local provides a Player backed by GStreamer for local audio files.
+// It plays file:// URIs directly so no DRM, no WebView, no credentials needed.
+package local
+
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/simone-vibes/vibez/internal/audioquality"
+	"github.com/simone-vibes/vibez/internal/player"
+	"github.com/simone-vibes/vibez/internal/player/gst"
+	"github.com/simone-vibes/vibez/internal/provider"
+)
+
+// Player implements player.Player for local audio files using GStreamer.
+type Player struct {
+	gst *gst.Player
+	mu sync.RWMutex
+	state player.State
+	subs []chan player.State
+	queue []provider.Track
+	idx int
+}
+
+// New creates a local Player backed by a GStreamer pipeline.
+func New() (*Player, error) {
+	gstPlayer, err := gst.New()
+	if err != nil {
+		return nil, fmt.Errorf("local player: %w", err)
+	}
+	p := &Player{gst: gstPlayer}
+	p.gst.OnEOS(func () {
+		_ = p.Next()
+	})
+	p.gst.OnError(func(e error) {
+		p.mu.Lock()
+		s := p.state
+		s.Error = e.Error()
+		p.mu.Unlock()
+		p.broadcast(s)
+	})
+	go p.pollState()
+	return p, nil
+}
+
+func (p *Player) broadcast(s player.State) {
+	p.mu.RLock()
+	defer p.mu.RUnlock()
+	for _, ch := range p.subs {
+		select  {
+		case ch <- s:
+		default:
+		}
+	}
+}
+
+func (p *Player) currentState() player.State {
+	p.mu.RLock()
+	s := p.state
+	p.mu.RUnlock()
+	s.Position = p.gst.Position()
+	return s
+}
+
+func (p *Player) playTrack(t provider.Track) {
+	uri := fmt.Sprintf("file://%s", t.ID[len("local:"):])
+	p.gst.PlayURI(uri)
+	time.Sleep(200 * time.Millisecond)
+	if d := p.gst.Duration(); d > 0 {
+		t.Duration = d
+	}
+	p.mu.Lock()
+	p.state.Track = &t
+	p.state.Playing = true
+	p.state.Position = 0
+	s := p.state
+	p.mu.Unlock()
+	p.broadcast(s)
+}
+
+func (p *Player) Play() error {
+	p.gst.Play()
+	p.mu.Lock()
+	p.state.Playing = true
+	s := p.state
+	p.mu.Unlock()
+	p.broadcast(s)
+	return nil
+}
+
+func (p *Player) Pause() error {
+	p.gst.Pause()
+	p.mu.Lock()
+	p.state.Playing = false
+	s := p.state
+	p.mu.Unlock()
+	p.broadcast(s)
+	return nil
+}
+
+func (p *Player) Stop() error {
+	p.gst.Stop()
+	p.mu.Lock()
+	p.state.Playing = false
+	s := p.state
+	p.mu.Unlock()
+	p.broadcast(s)
+	return nil
+}
+
+func (p *Player) Next() error {
+	p.mu.Lock()
+	if len(p.queue) == 0 {
+		p.mu.Unlock()
+		return nil
+	}
+	p.idx = (p.idx + 1) % len(p.queue)
+	t := p.queue[p.idx]
+	p.mu.Unlock()
+	p.playTrack(t)
+	return nil
+}
+
+func (p *Player) Previous() error {
+	p.mu.Lock()
+	if len(p.queue) == 0 {
+		p.mu.Unlock()
+		return nil
+	}
+	if p.gst.Position() > 3*time.Second {
+		p.mu.Unlock()
+		p.gst.Seek(0)
+		return nil
+	}
+	if p.idx == 0 {
+		p.idx = len(p.queue) - 1
+	} else {
+		p.idx--
+	}
+	t := p.queue[p.idx]
+	p.mu.Unlock()
+	p.playTrack(t)
+	return nil
+}
+
+func (p *Player) Seek(pos time.Duration) error {
+	p.gst.Seek(pos)
+	return nil
+}
+
+func (p *Player) SetVolume(v float64) error {
+	p.gst.SetVolume(v)
+	return nil
+}
+
+func (p *Player) SetAudioBitrate(kbps int) error {
+	if err := audioquality.Validate(kbps); err != nil {
+		return err
+	}
+	return player.ErrAudioBitrateSavedPreferenceOnly
+}
+
+func (p *Player) SetQueue(ids []string) error {
+	if len(ids) == 0 {
+		return nil
+	}
+	// Finding the first ID in the full queue and starting from there.
+	p.mu.Lock()
+	for i, t := range p.queue {
+		if t.ID == ids[0] {
+			p.idx = i
+			break
+		}
+	}
+	t := p.queue[p.idx]
+	p.mu.Unlock()
+	p.playTrack(t)
+	return nil
+}
+
+func (p *Player) SetPlaylist(_ string, startIdx int) error {
+	p.mu.Lock()
+	if len(p.queue) == 0 {
+		p.mu.Unlock()
+		return nil
+	}
+	if startIdx >= 0 && startIdx < len(p.queue) {
+		p.idx = startIdx
+	} else {
+		p.idx = 0
+	}
+	t := p.queue[p.idx]
+	p.mu.Unlock()
+	p.playTrack(t)
+	return nil
+}
+
+func (p *Player) AppendQueue(ids []string) error {
+	extra := tracksForIDs(p.queue, ids)
+	p.mu.Lock()
+	p.queue = append(p.queue, extra...)
+	p.mu.Unlock()
+	return nil
+}
+
+func (p *Player) SetRepeat(mode int) error {
+	p.mu.Lock()
+	p.state.RepeatMode = mode
+	p.mu.Unlock()
+	return nil
+}
+
+func (p *Player) SetShuffle(on bool) error {
+	p.mu.Lock()
+	p.state.ShuffleMode = on
+	p.mu.Unlock()
+	return nil
+}
+
+func (p *Player) SetEqualizer(_ []player.EQBand) error { return nil }
+
+func (p *Player) RemoveFromQueue(idx int) error {
+	p.mu.Lock()
+	if idx >= 0 && idx < len(p.queue) {
+		p.queue = append(p.queue[:idx], p.queue[idx+1:]...)
+	}
+	p.mu.Unlock()
+	return nil
+}
+
+func (p *Player) MoveInQueue(from, to int) error {
+	p.mu.Lock()
+	if from >= 0 && from < len(p.queue) && to >=0 && to < len(p.queue) {
+		t := p.queue[from]
+		p.queue = append(p.queue[:from], p.queue[from+1:]...)
+		p.queue = append(p.queue[:to], append([]provider.Track{t}, p.queue[to:]...)...)
+	}
+	p.mu.Unlock()
+	return nil
+}
+
+func (p *Player) ClearQueue() error {
+	p.gst.Stop()
+	p.mu.Lock()
+	p.queue = nil
+	p.idx = 0
+	p.state.Track = nil
+	p.state.Playing = false
+	s := p.state
+	p.mu.Unlock()
+	p.broadcast(s)
+	return nil
+}
+
+func (p *Player) GetState() (*player.State, error) {
+	s := p.currentState()
+	return &s,nil
+}
+
+func (p *Player) Subscribe() <- chan player.State {
+	ch := make(chan player.State, 8)
+	p.mu.Lock()
+	p.subs = append(p.subs, ch)
+	p.mu.Unlock()
+	return ch
+}
+
+func (p *Player) Close() error {
+	p.gst.Destroy()
+	return nil
+}
+
+// This replaces the internal queue with the given tracks. Called after
+// the provider scans the music directory
+func (p *Player) LoadTracks(tracks []provider.Track) {
+	p.mu.Lock()
+	p.queue = append([]provider.Track{}, tracks...)
+	p.idx = 0
+	p.mu.Unlock()
+}
+
+func tracksForIDs(tracks []provider.Track, ids []string) []provider.Track {
+	byID := make(map[string]provider.Track, len(tracks))
+	for _, t := range tracks {
+		byID[t.ID] = t
+	}
+	out := make([]provider.Track, 0, len(ids))
+	for _, id := range ids {
+		if t, ok := byID[id]; ok {
+			out = append(out, t)
+		}
+	}
+	return out
+}
+
+func (p *Player) pollState() {
+	ticker := time.NewTicker(500 * time.Millisecond)
+	defer ticker.Stop()
+	for range ticker.C {
+		p.mu.RLock()
+		playing := p.state.Playing
+		p.mu.RUnlock()
+		if !playing {
+			continue
+		}
+		p.mu.Lock()
+		pos := p.gst.Position()
+		p.state.Position = pos
+		s := p.state
+		p.mu.Unlock()
+		p.broadcast(s)
+	}
+}
--- a/internal/provider/local/provider.go
+++ b/internal/provider/local/provider.go
@@ -17,20 +17,20 @@ import (
 )
 
 var supportedExts = map[string]bool{
-	".mp3": true,
+	".mp3":  true,
 	".flac": true,
-	".m4a": true,
-	".ogg": true,
+	".m4a":  true,
+	".ogg":  true,
 }
 
 // Provider scans a local directory and exposes tracks via the provider interface.
 type Provider struct {
-	dir string
+	dir    string
 	tracks []provider.Track
 }
 
 // New creates a Provider and performs the initial directory scan.
-func New (dir string) (*Provider, error) {
+func New(dir string) (*Provider, error) {
 	p := &Provider{dir: dir}
 	if err := p.scan(); err != nil {
 		return nil, err
@@ -38,15 +38,15 @@ func New (dir string) (*Provider, error) {
 	return p, nil
 }
 
-func (p *Provider) Name() string {return "Local"}
-func (P *Provider) IsAuthenticated() bool {return true}
+func (p *Provider) Name() string          { return "Local" }
+func (p *Provider) IsAuthenticated() bool { return true }
 
 // scan walks dir recursively and indexes all supported audio files.
 func (p *Provider) scan() error {
 	p.tracks = nil
-	return filepath.Walk(p.dir, func(path string, info os.FileInfo ,err error) error {
+	return filepath.Walk(p.dir, func(path string, info os.FileInfo, err error) error {
 		if err != nil {
-			return nil
+			return err
 		}
 		if info.IsDir() {
 			return nil
@@ -66,11 +66,11 @@ func (p *Provider) scan() error {
 
 // trackFromFile reads metadata from an audio file and returns a Track.
 func trackFromFile(path string) (provider.Track, error) {
-	f, err := os.Open(path)
+	f, err := os.Open(path) //nolint:gosec // the path comes from user-configured music dir
 	if err != nil {
 		return provider.Track{}, err
 	}
-	defer f.Close()
+	defer func() { _ = f.Close() }()
 
 	m, err := tag.ReadFrom(f)
 
@@ -95,10 +95,10 @@ func trackFromFile(path string) (provider.Track, error) {
 	}
 
 	return provider.Track{
-		ID:	fmt.Sprintf("local:%s", path),
-		Title: title,
+		ID:     fmt.Sprintf("local:%s", path),
+		Title:  title,
 		Artist: artist,
-		Album: album,
+		Album:  album,
 		Genres: genres,
 	}, nil
 }
@@ -119,16 +119,16 @@ func (p *Provider) Search(_ context.Context, query string) (*provider.SearchResu
 		if strings.Contains(strings.ToLower(t.Title), q) ||
 			strings.Contains(strings.ToLower(t.Artist), q) ||
 			strings.Contains(strings.ToLower(t.Album), q) {
-				tracks = append(tracks, t)
-				if !seen[t.Album] {
-					seen[t.Album] = true
-					albums = append(albums, provider.Album{
-						ID: "local-album:" + t.Album,
-						Title:  t.Album,
-						Artist: t.Artist,
-					})
-				}
+			tracks = append(tracks, t)
+			if !seen[t.Album] {
+				seen[t.Album] = true
+				albums = append(albums, provider.Album{
+					ID:     "local-album:" + t.Album,
+					Title:  t.Album,
+					Artist: t.Artist,
+				})
 			}
+		}
 	}
 	return &provider.SearchResult{Tracks: tracks, Albums: albums}, nil
 }
@@ -149,7 +149,7 @@ func (p *Provider) GetAlbumTracks(_ context.Context, albumID string) ([]provider
 			out = append(out, t)
 		}
 	}
-	return out , nil
+	return out, nil
 }
 
 func (p *Provider) GetLibraryAlbumTracks(ctx context.Context, albumID string) ([]provider.Track, error) {
@@ -164,14 +164,14 @@ func (p *Provider) CreatePlaylist(_ context.Context, _ string, _ []string) (prov
 	return provider.Playlist{}, nil
 }
 
-func (p *Provider) LoveSong(_ context.Context, _ string, _ bool) error { return nil }
+func (p *Provider) LoveSong(_ context.Context, _ string, _ bool) error      { return nil }
 func (p *Provider) GetSongRating(_ context.Context, _ string) (bool, error) { return false, nil }
-func (p *Provider) AddToPlaylist(_ context.Context, _, _ string) error { return nil }
+func (p *Provider) AddToPlaylist(_ context.Context, _, _ string) error      { return nil }
 
 func (p *Provider) GetRecommendations(_ context.Context) ([]provider.RecommendationGroup, error) {
 	// Group tracks by genre using their metadata tags.
-	// Tracks with no genre tag are skipped.	
-	byGenre :=  map[string][]provider.Track{}
+	// Tracks with no genre tag are skipped.
+	byGenre := map[string][]provider.Track{}
 	for _, t := range p.tracks {
 		for _, g := range t.Genres {
 			if g != "" {
@@ -188,9 +188,9 @@ func (p *Provider) GetRecommendations(_ context.Context) ([]provider.Recommendat
 		items := make([]provider.RecommendationItem, 0, len(tracks))
 		for _, t := range tracks {
 			items = append(items, provider.RecommendationItem{
-				ID: t.ID,
-				Kind: "track",
-				Title: t.Title,
+				ID:       t.ID,
+				Kind:     "track",
+				Title:    t.Title,
 				Subtitle: t.Artist,
 			})
 		}
--- a/internal/provider/local/provider.go
+++ b/internal/provider/local/provider.go
@@ -1,0 +1,203 @@
+// Package local provides a Provider that serves tracks from a local music
+// directory. It scans recursively for MP3, FLAC, M4A and OGG files and reads
+// their metadata using the dhowden/tag library. No network or credentials
+// are required.
+
+package local
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/dhowden/tag"
+	"github.com/simone-vibes/vibez/internal/provider"
+)
+
+var supportedExts = map[string]bool{
+	".mp3": true,
+	".flac": true,
+	".m4a": true,
+	".ogg": true,
+}
+
+// Provider scans a local directory and exposes tracks via the provider interface.
+type Provider struct {
+	dir string
+	tracks []provider.Track
+}
+
+// New creates a Provider and performs the initial directory scan.
+func New (dir string) (*Provider, error) {
+	p := &Provider{dir: dir}
+	if err := p.scan(); err != nil {
+		return nil, err
+	}
+	return p, nil
+}
+
+func (p *Provider) Name() string {return "Local"}
+func (P *Provider) IsAuthenticated() bool {return true}
+
+// scan walks dir recursively and indexes all supported audio files.
+func (p *Provider) scan() error {
+	p.tracks = nil
+	return filepath.Walk(p.dir, func(path string, info os.FileInfo ,err error) error {
+		if err != nil {
+			return nil
+		}
+		if info.IsDir() {
+			return nil
+		}
+		ext := strings.ToLower(filepath.Ext(path))
+		if !supportedExts[ext] {
+			return nil
+		}
+		t, err := trackFromFile(path)
+		if err != nil {
+			return nil // skipping files that have unreadable metadata
+		}
+		p.tracks = append(p.tracks, t)
+		return nil
+	})
+}
+
+// trackFromFile reads metadata from an audio file and returns a Track.
+func trackFromFile(path string) (provider.Track, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return provider.Track{}, err
+	}
+	defer f.Close()
+
+	m, err := tag.ReadFrom(f)
+
+	title := filepath.Base(path)
+	artist := "Unknown Artist"
+	album := "Unknown Album"
+	var genres []string
+
+	if err == nil {
+		if m.Title() != "" {
+			title = m.Title()
+		}
+		if m.Artist() != "" {
+			artist = m.Artist()
+		}
+		if m.Album() != "" {
+			album = m.Album()
+		}
+		if m.Genre() != "" {
+			genres = []string{m.Genre()}
+		}
+	}
+
+	return provider.Track{
+		ID:	fmt.Sprintf("local:%s", path),
+		Title: title,
+		Artist: artist,
+		Album: album,
+		Genres: genres,
+	}, nil
+}
+
+func (p *Provider) GetLibraryTracks(_ context.Context) ([]provider.Track, error) {
+	out := make([]provider.Track, len(p.tracks))
+	copy(out, p.tracks)
+	return out, nil
+}
+
+func (p *Provider) Search(_ context.Context, query string) (*provider.SearchResult, error) {
+	q := strings.ToLower(query)
+	var tracks []provider.Track
+	var albums []provider.Album
+	seen := map[string]bool{}
+
+	for _, t := range p.tracks {
+		if strings.Contains(strings.ToLower(t.Title), q) ||
+			strings.Contains(strings.ToLower(t.Artist), q) ||
+			strings.Contains(strings.ToLower(t.Album), q) {
+				tracks = append(tracks, t)
+				if !seen[t.Album] {
+					seen[t.Album] = true
+					albums = append(albums, provider.Album{
+						ID: "local-album:" + t.Album,
+						Title:  t.Album,
+						Artist: t.Artist,
+					})
+				}
+			}
+	}
+	return &provider.SearchResult{Tracks: tracks, Albums: albums}, nil
+}
+
+func (p *Provider) GetLibraryPlaylists(_ context.Context) ([]provider.Playlist, error) {
+	return nil, nil
+}
+
+func (p *Provider) GetPlaylistTracks(_ context.Context, _ string) ([]provider.Track, error) {
+	return nil, nil
+}
+
+func (p *Provider) GetAlbumTracks(_ context.Context, albumID string) ([]provider.Track, error) {
+	albumID = strings.TrimPrefix(albumID, "local-album:")
+	var out []provider.Track
+	for _, t := range p.tracks {
+		if t.Album == albumID {
+			out = append(out, t)
+		}
+	}
+	return out , nil
+}
+
+func (p *Provider) GetLibraryAlbumTracks(ctx context.Context, albumID string) ([]provider.Track, error) {
+	return p.GetAlbumTracks(ctx, albumID)
+}
+
+func (p *Provider) GetCatalogPlaylistTracks(_ context.Context, _ string) ([]provider.Track, error) {
+	return nil, nil
+}
+
+func (p *Provider) CreatePlaylist(_ context.Context, _ string, _ []string) (provider.Playlist, error) {
+	return provider.Playlist{}, nil
+}
+
+func (p *Provider) LoveSong(_ context.Context, _ string, _ bool) error { return nil }
+func (p *Provider) GetSongRating(_ context.Context, _ string) (bool, error) { return false, nil }
+func (p *Provider) AddToPlaylist(_ context.Context, _, _ string) error { return nil }
+
+func (p *Provider) GetRecommendations(_ context.Context) ([]provider.RecommendationGroup, error) {
+	// Group tracks by genre using their metadata tags.
+	// Tracks with no genre tag are skipped.	
+	byGenre :=  map[string][]provider.Track{}
+	for _, t := range p.tracks {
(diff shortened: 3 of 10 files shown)


internal/provider/local/provider.go
```diff
@@ -0,0 +1,203 @@
...
+
+func (p *Provider) Name() string {return "Local"}
+func (P *Provider) IsAuthenticated() bool {return true}
+
+// scan walks dir recursively and indexes all supported audio files.
+func (p *Provider) scan() error {
+	p.tracks = nil
+	return filepath.Walk(p.dir, func(path string, info os.FileInfo ,err error) error {
+		if err != nil {
+			return nil
```
simonepelosi: Is this silencing the errors ? If yes, this means the tests are passing but they are not catching what we want

internal/player/local/player.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+	// Finding the first ID in the full queue and starting from there.
+	p.mu.Lock()
+	for i, t := range p.queue {
+		if t.ID == ids[0] {
+			p.idx = i
+			break
+		}
+	}
+	t := p.queue[p.idx]
```
simonepelosi: If the requested ID is absent from the queue p.idx keeps its old value and the wrong track plays silently. If p.queue is empty (called before LoadTracks) the slice access panics. Track a found bool and return
 early when not found.

internal/player/local/player.go
```diff
@@ -0,0 +1,316 @@
...
+	p := &Player{gst: gstPlayer}
+	p.gst.OnEOS(func () {
+		_ = p.Next()
+	})
+	p.gst.OnError(func(e error) {
+		p.mu.Lock()
+		s := p.state
+		s.Error = e.Error()
+		p.mu.Unlock()
+		p.broadcast(s)
```
simonepelosi, line 45: p.state.Error is never set. The broadcast carries the error once, but the next pollState tick re-broadcasts a clean state. Fix: assign p.state.Error = e.Error() inside the lock before copying.

   s := p.state        // ← copy
   s.Error = e.Error() // ← mutates the copy, not p.state

internal/player/local/player.go
```diff
@@ -0,0 +1,316 @@
+//go:build linux
```
simonepelosi, line 1: What about MacOS?

internal/player/local/player.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+	p.mu.Unlock()
+	return nil
+}
+
+func (p *Player) MoveInQueue(from, to int) error {
+	p.mu.Lock()
+	if from >= 0 && from < len(p.queue) && to >=0 && to < len(p.queue) {
+		t := p.queue[from]
+		p.queue = append(p.queue[:from], p.queue[from+1:]...)
```
simonepelosi, line 263:
```go
   // lines 237-239
   t := p.queue[from]
   p.queue = append(p.queue[:from], p.queue[from+1:]...)           // ← all elements after `from` shift left by 1
   p.queue = append(p.queue[:to], append([]provider.Track{t}, p.queue[to:]...)...)  // ← `to` not adjusted
 ```

 After line 238 removes the element at from, every element with index > from shifts left by one. When from < to, position to in the shrunken slice corresponds to to - 1 in the original. The current code inserts one position too far right. Fix: add if from < to { to-- } between lines 238 and 239.

File: cmd/root.go [hunk 6]
```diff
@@ -69,6 +76,62 @@ func runTUI(_ *cobra.Command, _ []string) error {
...
+				fmt.Fprintf(os.Stderr, "debug: saving music dir: %v\n", saveErr)
+			}
+		}
+
+		// If musicDir is still empty, prompt the user on the command line
+		// before launching the TUI.
+		if cfg.MusicDir == "" {
+			fmt.Print("Enter path to your music directory: ")
+			var dir string
+			fmt.Scanln(&dir)
```
simonepelosi:
```go
   // lines 92-93
   var dir string
   fmt.Scanln(&dir)  // ← stops at first space; "/home/user/My Music" → "/home/user/My"
 ```

 Use bufio.NewReader(os.Stdin).ReadString('\n') followed by strings.TrimSpace.

File: internal/player/local/player.go [hunk 7]
```diff
@@ -0,0 +1,316 @@
...
+	defer ticker.Stop()
+	for range ticker.C {
+		p.mu.RLock()
+		playing := p.state.Playing
+		p.mu.RUnlock()
+		if !playing {
+			continue
+		}
+		p.mu.Lock()
+		pos := p.gst.Position()
```
simonepelosi: Add a done chan struct{} field, close it in Close(), and change the loop:

 ```go
   for {
       select {
       case <-ticker.C:
           // ... existing poll logic
       case <-p.done:
           return
       }
   }
 ```

simonepelosi (changes requested):
Thank you so much for  this important contribution! I tried to review everything! Please run also linters and tests!

GurKalra:
> Thank you so much for this important contribution! I tried to review everything! Please run also linters and tests!

Ran the linters and the tests fully!

internal/provider/local/provider.go
GurKalra: Fixed in the latest push!!

internal/player/local/player.go
GurKalra: Fixed in the latest push!!

internal/player/local/player.go
GurKalra: Fixed in the latest push!!

internal/player/local/player.go
GurKalra: For macOS support I researched a few options. The proper native approach would be AVFoundation or CoreAudio directly via CGo, which would work great but is significant work. I also found github.com/gen2brain/malgo which is Go bindings for miniaudio. It is a single C library that wraps CoreAudio on macOS and ALSA/PulseAudio on Linux, no external dependencies needed on either platform. It could potentially replace GStreamer entirely in a future.

I also found github.com/ebitengine/oto which handles audio output on both Linux and macOS with no CGo, but it only outputs raw PCM bytes so I'd need separate decoder libraries for each format (MP3, FLAC, M4A, OGG) which would turn into dependency hell. 
 
However I don't have a Mac to test on, so whatever I write for macOS I can't actually verify works. 
What do you think is the best move forward, let me know?

internal/player/local/player.go
GurKalra: Fixed in the latest push!

cmd/root.go
GurKalra: Fixed in the latest push!

internal/player/local/player.go
GurKalra: Fixed in the latest push!

simonepelosi:
Hey! Thank you again for your great contribution! GH actions are still unhappy, can you double check ?

GurKalra:
> Hey! Thank you again for your great contribution! GH actions are still unhappy, can you double check ?

that is because i haven't written anything for the mac-vet-test. I have already written in the issue that you raised asking me about the mac test but lemme write it here again as well.

For macOS support I researched a few options. The proper native approach would be AVFoundation or CoreAudio directly via CGo, which would work great but is significant work. I also found github.com/gen2brain/malgo which is Go bindings for miniaudio. It is a single C library that wraps CoreAudio on macOS and ALSA/PulseAudio on Linux, no external dependencies needed on either platform. It could potentially replace GStreamer entirely in a future.

I also found github.com/ebitengine/oto which handles audio output on both Linux and macOS with no CGo, but it only outputs raw PCM bytes so I'd need separate decoder libraries for each format (MP3, FLAC, M4A, OGG) which would turn into dependency hell.

However I don't have a Mac to test on, so whatever I write for macOS I can't actually verify works.
What do you think is the best move forward, let me know?

simonepelosi:
Sorry for the late reply but it was a busy week, thank you for your great work. I would love to see something with CoreAudio and I can test it on my side. Do you think you have time to work on that ?

GurKalra:
> Sorry for the late reply but it was a busy week, thank you for your great work. I would love to see something with CoreAudio and I can test it on my side. Do you think you have time to work on that ?

Sorry for such a late reply!! Definitely this will be possible, i will get to it right away. Again thanks a lot and sorry for the late reply!

simonepelosi:
@GurKalra No worries at all! I really appreciate what you are doing! Please ping me when the code is ready.

ianswope:
@GurKalra I went through this while reading over the open queue. Hopefully some of it saves you time when you pick it back up, no rush intended.

Worth noting since the review still shows as changes-requested: all six of Simone's inline points did land in 0e610c4. The stdin read is now `bufio.ReadString` (root.go:94), the poll goroutine has a `done` channel (player.go:26, 282, 327), `state.Error` is assigned before the broadcast (player.go:41), the ID lookup tracks `found` and guards the empty queue (player.go:173-181), and the `MoveInQueue` arithmetic is correct now with the `if from < to { to-- }` adjustment.

CI is red on three mechanical things rather than anything structural: gofmt on `provider_test.go:201`, the `go mod tidy` hook, and the end-of-file fixer. The macOS job is the separate CoreAudio gap you and Simone already discussed.

Two things I noticed reading through that I don't think have come up yet.

`p.idx` isn't adjusted when the queue is mutated. `RemoveFromQueue` (player.go:233) and `MoveInQueue` (player.go:242) both reslice `p.queue` while `p.idx` stays put, so removing or moving a track above the current position leaves the current index pointing at a different track than the one actually playing. Removing the playing track doesn't stop or advance playback either.

`scan()` returns the `filepath.Walk` error directly (provider.go:47-49), so a single unreadable subdirectory anywhere under the music dir aborts the whole scan and `New()` fails, so local mode won't start at all. Returning nil there (or `filepath.SkipDir`) would let it continue. The per-file handling just below at provider.go:59 already does the lenient thing. I suspect this is what Simone was getting at with the question at provider.go:49: files are skipped correctly, it's the directory level that's too strict.

Rebase looks light whenever you get there. 25 commits of drift, but the only conflicts are `go.mod` and `go.sum`; `model.go` merges cleanly.

Happy to take another look once you push.

GurKalra:
Hi @simonepelosi @ianswope !! CoreAudio implementation is ready for testing! Added internal/player/local/player_darwin.go using ExtAudioFile for decoding and AudioQueue for playback. Both are part of AudioToolbox which ships with every Mac so no extra dependencies needed. Would love to hear if it works on your end!

ianswope:
@GurKalra Thanks for picking this back up, and for taking the `p.idx` and `scan()` notes on board.

I'm traveling with only a Linux box until Friday, so I can't test the darwin path before then — happy to run it on my Mac once I'm home.

One thing first: `player_darwin.go` doesn't compile on darwin yet. Linux is unaffected, since that file is excluded there, which is probably why it looks clean locally.

- `:449` — `p.queue...[to:]...` doesn't parse; wants `p.queue[to:]...`
- `:423` — `p.queue[idx+1]...` is missing the slice bound
- `:418` — `Player.EQBand` should be `player.EQBand`
- `:404` — `tracksForIDs` lives in `player.go`, which is `//go:build linux`, so it's undefined on darwin
- `:151`, `:157`, `:161` — unused imports; the golangci-lint and x/exp ones are what pulled golangci-lint, viper and testify into `go.mod`, so dropping them and running `go mod tidy` should shrink that diff a lot

`gofmt -e` catches :449 from Linux, but not the others — `GOOS=darwin go build` implies `CGO_ENABLED=0`, which skips any file importing `"C"`.

Which is the bigger issue, actually: the `build-vet-test-macos` job builds with `CGO_ENABLED=0`, so on darwin it excludes `player_darwin.go`, `player.go` is linux-only, and the package ends up with no files at all. That job can't go green with a cgo-based player as things stand — it needs a `darwin && !cgo` stub alongside it, or cgo enabled in CI. Worth settling with @simonepelosi before more work goes in, since it shapes the file layout.

Also, CI hasn't run since `0e610c4` — I think the merge conflict is stopping it from queuing, so a rebase should get it flowing again.

I have a handful of runtime notes as well (EOS never reaches Go, so playback looks like it stops at the end of a track rather than advancing; and `CFURLCreateWithString` returns NULL for any filename with a space, unchecked). I'll go through those properly once it builds — ping me when it does.

GurKalra:
> @GurKalra Thanks for picking this back up, and for taking the `p.idx` and `scan()` notes on board.
> 
> I'm traveling with only a Linux box until Friday, so I can't test the darwin path before then — happy to run it on my Mac once I'm home.
> 
> One thing first: `player_darwin.go` doesn't compile on darwin yet. Linux is unaffected, since that file is excluded there, which is probably why it looks clean locally.
> 
> * `:449` — `p.queue...[to:]...` doesn't parse; wants `p.queue[to:]...`
> * `:423` — `p.queue[idx+1]...` is missing the slice bound
> * `:418` — `Player.EQBand` should be `player.EQBand`
> * `:404` — `tracksForIDs` lives in `player.go`, which is `//go:build linux`, so it's undefined on darwin
> * `:151`, `:157`, `:161` — unused imports; the golangci-lint and x/exp ones are what pulled golangci-lint, viper and testify into `go.mod`, so dropping them and running `go mod tidy` should shrink that diff a lot
> 
> `gofmt -e` catches :449 from Linux, but not the others — `GOOS=darwin go build` implies `CGO_ENABLED=0`, which skips any file importing `"C"`.
> 
> Which is the bigger issue, actually: the `build-vet-test-macos` job builds with `CGO_ENABLED=0`, so on darwin it excludes `player_darwin.go`, `player.go` is linux-only, and the package ends up with no files at all. That job can't go green with a cgo-based player as things stand — it needs a `darwin && !cgo` stub alongside it, or cgo enabled in CI. Worth settling with @simonepelosi before more work goes in, since it shapes the file layout.
> 
> Also, CI hasn't run since `0e610c4` — I think the merge conflict is stopping it from queuing, so a rebase should get it flowing again.
> 
> I have a handful of runtime notes as well (EOS never reaches Go, so playback looks like it stops at the end of a track rather than advancing; and `CFURLCreateWithString` returns NULL for any filename with a space, unchecked). I'll go through those properly once it builds — ping me when it does.

Fixed all the compile errors  
- slice syntax typos 
- player.EQBand
- copied tracksForIDs into the darwin file
- cleaned up go.mod. 

The CGo/CI question is still open. Should I enable CGo in the macOS CI job or add a //go:build darwin && !cgo stub alongside the real player? @simonepelosi

GurKalra:
Rebased onto the latest upstream main. Picked up the go-jose v3.0.5 bump, the playwright-go upgrade, and the new GetStationTracks interface method (stubbed as nil, nil for local since there are no radio stations in a local library).

Summary of everything that's been addressed since the last review:

- All six of Simone's original inline comments fixed
- CI mechanical issues resolved (gofmt, goimports, end-of-file, go mod tidy)
- p.idx now adjusted correctly in RemoveFromQueue and MoveInQueue
- scan() skips unreadable subdirectories instead of failing the whole scan
- CoreAudio player added in player_darwin.go using ExtAudioFile + AudioQueue
- Darwin compile errors fixed (slice syntax, player.EQBand, tracksForIDs copied)
- CGo/macOS CI question still open from my earlier comment !!

CI is waiting for workflow approval. Ready for another look when you get a chance!

simonepelosi:
Thanks for picking this up and addressing the earlier feedback, @GurKalra! And thank you so much @ianswope for your guidance.

A few things needed before we can get this merged:

- Build Error (GetStationTracks): go test ./... fails right now because internal/provider/local/provider.go doesn't implement the GetStationTracks method on provider.Provider. Adding this stub to provider.go fixes the build.

- Rebase onto main: main moved ahead recently (including adding Warnings []string to provider.SearchResult). A quick rebase on main will keep everything in sync.

- macOS CGo & CI Strategy: To answer your question regarding CGo in CI:
I would add a player_darwin_nocgo.go file with //go:build darwin && !cgo containing a stub New() function (returning fmt.Errorf("local playback requires CGo on macOS")). That way, CGO_ENABLED=0 builds on macOS won't fail with "no Go files in package".
Then I can update the GitHub Actions workflow so the macOS runner builds with CGO_ENABLED=1 to exercise the real CoreAudio player.

- CoreAudio Details (player_darwin.go): Track auto-advancing on EOS: Inside vibez_audio_callback (line 46), AudioQueueStop stops output on end-of-file, but doesn't notify Go to advance to the next track. Calling p.Next() on EOS will keep queue playback going continuously.

- Paths with spaces: Lines 63–64 use CFURLCreateWithString, which returns NULL when file paths contain spaces or special characters. Switching to CFURLCreateWithFileSystemPath(NULL, cfPath, kCFURLPOSIXPathStyle, false) (or URL-encoding the path) ensures songs with spaces in their path open correctly.

ianswope:
@GurKalra Thanks for the fast turnaround — c348e518 covers all of Simone's actionable points as far as I can tell from reading it: the `GetStationTracks` stub (provider.go:208), the `darwin && !cgo` stub, `CFURLCreateWithFileSystemPath` replacing `CFURLCreateWithString`, and EOS calling back into Go. The shape is right; what's left is mechanical.

Two of them will stop CI, so worth pushing before anyone spends a workflow approval on the run:

**`fmt.Error` doesn't exist** — `player_darwin_nocgo.go` needs `fmt.Errorf`. This one matters more than it looks: the macOS job's first step is `CGO_ENABLED=0 go build ./...` (ci.yml:76), which is exactly the configuration that compiles this stub, so the run fails here in seconds.

**gofmt** — `func New() (*Player, error){` needs a space before the brace, or `pre-commit` fails.

Then two in the new EOS path in `player_darwin.go`:

**`// export vibezOnEOS` doesn't export anything.** The directive has to be `//export vibezOnEOS`, no space after the slashes. With the space it's an ordinary comment, cgo never emits the symbol, and the `extern void vibezOnEOS(void*)` in the preamble stays undefined at link time.

**A Go pointer is stored in C memory.** `audio.goPlayer = unsafe.Pointer(p)` keeps a Go pointer in a C struct that outlives the call, which the cgo pointer rules don't permit. `runtime/cgo.Handle` is the supported route: `h := cgo.NewHandle(p)`, store the handle, recover with `h.Value().(*Player)`, `h.Delete()` when the stream closes.

**`p.Next()` runs on the audio render thread.** `vibezOnEOS` fires from inside `vibez_audio_callback`, so `Next()` — which takes `p.mu` and does provider work — runs on a realtime callback thread that may already hold that lock via the calling path. Signalling a channel for a goroutine to drain keeps the callback non-blocking and avoids the deadlock risk.

@simonepelosi one thing worth knowing before the next approval: no CI job currently links the cgo player. `internal/player/local` has no test files, so `go test ./...` never links it, and both macOS build steps pin `CGO_ENABLED=0`, which selects the stub instead. So the `//export` problem above would pass CI silently — it only shows up in a real cgo darwin build. That's a good reason for the `CGO_ENABLED=1` macOS job you suggested; it would have caught it.

I haven't compiled any of this — it's from reading the patch, so treat the last two as review notes rather than confirmed failures. I'll run a real darwin build and playback test and report back.

ianswope:
@GurKalra Thanks for turning this around so fast. The shape of all five is right — `cgo.Handle` in place of the raw pointer, the `//export` directive fixed, and the EOS path now doing a non-blocking send that a separate goroutine drains, so `Next()` is off the render thread entirely. That last one is exactly what I was hoping for.

Two things will stop a real darwin build, though, and neither is caught by CI for the reason below.

**`vibezOnEOS` doesn't compile.** The parameter is named `ptr` but the body reads `h`:

```go
func vibezOnEOS(ptr unsafe.Pointer) {   // player_darwin.go:235
	handle := cgo.Handle(h)             // :236 — undefined: h
```

Related, and worth fixing together: the preamble declares `extern void vibezOnEOS(uintptr_t handle)` (`:29`), but an exported Go func taking `unsafe.Pointer` makes cgo emit `void vibezOnEOS(void*)` into `_cgo_export.h`, so the two declarations conflict. Taking the handle as an integer end to end lines them up:

```go
//export vibezOnEOS
func vibezOnEOS(h C.uintptr_t) {
	p := cgo.Handle(uintptr(h)).Value().(*Player)
	...
}
```

**gofmt fails on `player_darwin.go`.** `gofmt -l` flags it — the two new struct fields change the alignment of the whole `Player` block, `done:` in the `New()` literal needs to shift with `eosCh:`, and `eosLoop()`/`for`/`select` each want a space before the brace. `gofmt -w internal/player/local/player_darwin.go` covers all of it. `player_darwin_nocgo.go` is clean now. This one does fail CI — `pre-commit` runs regardless of platform.

One suggestion rather than a finding, since I haven't run vet on it and it may well be fine as written: if you're changing the signature to `C.uintptr_t` anyway, carrying the handle as an integer into the C struct too — `uintptr_t goPlayer`, assigned with `C.uintptr_t(p.handle)` at `:274` — would keep the types consistent with the `extern` and avoid the `uintptr`→`unsafe.Pointer` conversion, which is the kind of thing `unsafeptr` sometimes has opinions about.

@simonepelosi this is the same blind spot from my last comment: `internal/player/local` has no test files, so `go test ./...` never links the cgo player, and both macOS steps pin `CGO_ENABLED=0` and select the stub. So the undefined `h` above — an outright compile error — passes CI clean today. The `CGO_ENABLED=1` macOS job would have caught it in seconds, and I think it's worth landing before more darwin code goes in.

I ran `gofmt` on the pushed files, but I haven't done a real cgo darwin build — I don't have the Mac in front of me yet. Once I do I'll build for real and go through the runtime side properly, including whether EOS now advances the queue end to end and the space-in-path fix. Ping me when the two above are pushed.

ianswope:
Thanks for sticking with this — the CoreAudio backend is a real chunk of work, and I want to lead with the good news: **the EOS callback path works.** I got it running on hardware and auto-advance fires cleanly. Details below.

I tested this on an **M-series Mac, macOS 25.5, Apple clang 21.0.0, Go darwin/arm64**, at head `b306c6a`. Test corpus was six 5.000s sine files (mp3, flac, m4a, ogg, one 48 kHz flac, one with `#` in the filename), all tagged.

---

## Build: both paths still fail at `b306c6a`

`b306c6a`'s message says it fixed the `vibezOnEOS` parameter-name mismatch and moved to `C.uintptr_t`. Only the function **body** was changed — the signature line above it is untouched, so the C and Go halves now disagree:

```
$ CGO_ENABLED=1 go build ./...
# github.com/simone-vibes/vibez/internal/player/local
In file included from _cgo_export.c:4:
cgo-gcc-export-header-prolog:55:13: error: conflicting types for 'vibezOnEOS'
player_darwin.go:29:13: note: previous declaration is here
_cgo_export.c:26:6: error: conflicting types for 'vibezOnEOS'
```

The preamble declares `extern void vibezOnEOS(uintptr_t)`, but the Go side is still `func vibezOnEOS(ptr unsafe.Pointer)`, which cgo emits as `void vibezOnEOS(void*)`. Either side can move, they just have to match.

Behind that sits a second error: the body reads `h`, but the parameter is named `ptr`.

Once the signature is consistent, the next failure is:

```
cmd/root.go:132:45: ... does not implement player.Player (missing method SetShuffle)
```

`SetShuffle` is the **only** interface method the Linux GStreamer player has that the CoreAudio one lacks — I diffed all 25 methods across both files.

And the no-cgo path fails separately:

```
$ CGO_ENABLED=0 go build ./...
cmd/root.go:120:8: plyr.LoadTracks undefined
cmd/root.go:127:13: plyr.SetQueue undefined
cmd/root.go:132:45: ... missing method AppendQueue
```

`player_darwin_nocgo.go` is 11 lines — `type Player struct{}` plus a `New()` that returns the "requires CGo" error. It needs the rest of the interface surface (or `cmd/root.go` needs to stop reaching for the concrete type), otherwise `CGO_ENABLED=0` can't build on macOS at all.

Worth noting no CI has ever run on this branch, so none of the above was caught automatically.

---

## Runtime: what I found after getting it to build

To get past the compile errors I applied two throwaway local edits — corrected the `vibezOnEOS` signature and added a one-line `SetShuffle`. **These are my scaffolding, not proposed code**; I'd rather you make the calls on your own implementation. Everything below is from the resulting binary, driving the `Player` API directly (volume forced to 0).

### ✅ EOS auto-advance works

```
t=0.0s  track=TRACK-A
t=4.5s  track=TRACK-A  pos=5.015510204s  playing=true
t=5.0s  track=TRACK-B  pos=557.278911ms  playing=true
RESULT: EOS auto-advance FIRED after 5.0s (TRACK-A -> TRACK-B)
```

The C callback → `cgo.Handle` → `eosCh` → `eosLoop` → `Next()` chain does what it's meant to. Nice.

### 1. `Seek` kills playback permanently

Seeking mid-playback (no EOS involved) stops audio for good, while `Playing` stays `true` so the UI keeps claiming it's playing:

```
t=2.0s (before seek):     pos=2.414875283s
t=+0.7s after Seek(1s):   pos=1s  playing=true
t=+1.4s after Seek(1s):   pos=1s  playing=true
t=+2.1s after Seek(1s):   pos=1s  playing=true
t=+2.8s after Seek(1s):   pos=1s  playing=true
```

The read cursor never advances past the seek target. `vibez_seek` does `AudioQueueStop(queue, true)` then `AudioQueueStart` without re-allocating or re-enqueueing any buffers — an immediate stop returns all buffers, so there's nothing left in the queue for the output callback to drain, and it's never invoked again. This is the highest-impact item on the list; seek is a core control.

### 2. Sample rate is hardcoded to 44100 in three places

`vibez_get_duration` returns `kExtAudioFileProperty_FileLengthFrames`, which is in the **file's** frames, but duration/position/seek all divide or multiply by a literal `44100`. A 48 kHz file is misreported by exactly the ratio:

```
48kHz flac  file=six_48k.flac  duration=5.44217687s   (actual 5.000s)
```

`5 × 48000/44100 = 5.44217687`. Same literal is in `Seek` (`pos.Seconds() * 44100`) and `GetState`, so seek targets and the progress bar drift on anything that isn't 44.1 kHz. The client format you set is 44.1 kHz, but the file length and `ExtAudioFileTell` are not in client frames — worth reconciling with `format.mSampleRate` or converting explicitly.

### 3. Position is the decode cursor, not the playback position

Every track's position starts at **557.278911ms**, never 0, and stays ~0.55s ahead of what you'd hear — visible above, where it reads `5.015s` while a 5.000s track is still playing. That's exactly the prefill: 3 buffers × 32768 bytes ÷ 4 bytes/frame = 24576 frames ÷ 44100 = 557.3ms. `ExtAudioFileTell` reports how far *decoding* has got, not playback. `AudioQueueGetCurrentTime` is the usual source for this.

### 4. `.ogg` is listed but unplayable on macOS

The provider's extension map includes `.ogg`, and the PR description and ROADMAP both advertise OGG. But ExtAudioFile has no Vorbis decoder on macOS, so the file opens, yields zero frames, immediately reports EOS, and gets skipped in under a second:

```
t= 0.8s track=OGG       pos=0s  dur=0s  playing=true
t= 1.5s track=NEXT-MP3  pos=1.114557823s  dur=5.041632653s
RESULT: advanced off the ogg after 1.5s
```

So on macOS an ogg-heavy library silently flashes past, one file per second, with nothing in the UI explaining why. My suggestion is to scope OGG to Linux rather than chase a decoder: GStreamer handles Vorbis on the Linux path, so this is a genuine platform capability difference, and bundling a Vorbis decoder would be a big addition to a PR this size. Bear in mind `internal/provider/local/provider.go` has no build tag, so `supportedExts` is currently shared — scoping it means splitting that set per platform, and the OGG claim is also repeated in that file's package doc comment, in this PR's description, and in ROADMAP.md.

(mp3, flac and m4a all scan, tag and play correctly.)

### 5. Single-track queue loops forever, ignoring repeat mode

With `SetRepeat(0)` (`RepeatModeOff`), a one-track queue restarted 3 times in 12 seconds:

```
t= 0.8s pos WRAPPED to 1.30031746s  (restart #1)
t= 5.3s pos WRAPPED to 928.798185ms (restart #2)
t= 9.8s pos WRAPPED to 557.278911ms (restart #3)
```

`Next()` wraps unconditionally via `p.idx = (p.idx + 1) % len(p.queue)`, and `SetRepeat` stores the mode without anything ever consulting it.

### 6. `SetQueue` ignores its ID list

The interface contract is "replaces the playback queue with the given catalog track IDs". The implementation searches the *existing* queue for `ids[0]` and just jumps to it:

```
SetQueue([C,B]) -> now playing="TRACK-C" queue len=3 idx=2   (want len=2)
```

So queueing a subset of search results leaves the queue and the queue panel disagreeing about what comes next.

### 7. The queue is unrecoverable after `ClearQueue`

```
after ClearQueue + SetQueue([A]) + AppendQueue([A,B]): queue len=0 track="" playing=false
```

`ClearQueue` sets `p.queue = nil`; `AppendQueue` resolves incoming IDs against `p.queue` through `tracksForIDs`, so once it's empty nothing can ever be added back, and `SetQueue` returns early for the same reason. Practically: `:clear` bricks playback until restart.

### 8. `pollState` broadcasts a stale position

The Linux player refreshes position inside its poll loop (`p.state.Position = p.gst.Position()`). The darwin `pollState` re-broadcasts `p.state` without refreshing it — only `GetState` recomputes. Subscribers (TUI progress, and MPRIS on the Linux side) would see a frozen position.

---

## Things I could not reproduce

I suspected a panic from `playTrack` doing `handle.Delete()` + `cgo.NewHandle()` outside the mutex while an EOS callback is in flight. **It did not reproduce** across 300 rapid `Next()`/`Previous()` transitions, 6 deliberate collisions at natural EOS boundaries, and `Close()` called 100ms before EOS. Flagging the pattern as worth a second look rather than reporting it as a bug — I couldn't make it crash.

Also fine: a `#` in the filename plays without trouble (raw POSIX path, no `file://` round-trip), and `go test ./internal/provider/local/` passes on darwin.

---

## Summary

The architecture is sound and the hard part — format-agnostic decode plus a working EOS callback — is done. That's genuinely the part I'd have expected to go wrong, and it doesn't.

I know this is a long list, so here's the order I'd tackle it in rather than all at once:

1. **The four compile errors** — nothing else can be verified until both build paths work. Get CI green on the branch at the same time; every one of these would have been caught there.
2. **Seek (#1)** — the only item that makes a core control non-functional.
3. **The 44100 literals (#2, #3)** — one root cause behind both the wrong durations and the drifting progress bar.
4. **Queue contract gaps (#6, #7)** — `SetQueue` ignoring its ID list, and `:clear` bricking playback.
5. **Repeat mode (#5) and stale poll state (#8)** — smaller, mechanical.
6. **OGG (#4)** — a scoping decision more than a bug; worth settling but not worth blocking on.

None of this needs a Mac from you. Happy to re-run the whole matrix on Apple Silicon whenever you push — the harness exists now, so turnaround is a few minutes, and I'll report back with the same detail.

simonepelosi:
Please can you solve conflicts ? @GurKalra

GurKalra:
> Please can you solve conflicts ? @GurKalra

Done!!

simonepelosi:
@GurKalra sorry for the long back and forth, the macos pipeline is failing, can you double check it ?

GurKalra:
> @GurKalra sorry for the long back and forth, the macos pipeline is failing, can you double check it ?

Fixed the macOS test failure. TestProvider_GetLibroadTracks_OnlyAudioFiles was expecting 4 tracks including .ogg on all platforms, but we scoped OGG to Linux only since ExtAudioFile has no Vorbis decoder on macOS. Added platform-specific wantAudioFileCount constants so the test expects 4 on Linux and 3 on macOS. Also rebased onto latest main which added install/uninstall Makefile targets.

ianswope:
CI hasn't run on `91a1d14`. The run from that push (`31504103360`) is sitting at `action_required` with a 0s duration — @simonepelosi it needs an approve-and-run before it reports anything. Four earlier pushes are in the same state; the only runs that ever executed are the ones that got approved. So the macOS question is still open in both directions.

@GurKalra one thing worth fixing before that run happens, or it comes back red on a different job. `gofmt` fails on three of the four new files:

- `internal/provider/local/exts_darwin.go` — map values need aligning (`".mp3":  true`, `".m4a":  true`)
- `internal/provider/local/exts_darwin_test.go` — no trailing newline
- `internal/provider/local/exts_linux_test.go` — no trailing newline

`gofmt` doesn't honour build tags, so the `pre-commit` job checks the darwin-tagged files on Linux too. `gofmt -w internal/provider/local/` covers all three. Worth landing in the same push — otherwise the approved run goes red on `pre-commit` and reads as still broken.

The scoping itself is right. Build-tagged `supportedExts` with a matching per-platform `wantAudioFileCount` is cleaner than branching inside the test, and `vibezOnEOS` now matches its body (`C.uintptr_t` in, `cgo.Handle(uintptr(h))` out) — that one's resolved.

ianswope:
@simonepelosi separately, something that isn't GurKalra's to fix and probably needs your call before this lands.

`.goreleaser.yml:38` builds the darwin binaries with `CGO_ENABLED=0`, and the release job runs on `ubuntu-latest`. The nocgo stub's `New()` returns `local playback requires CGo on macOS — build with CGO_ENABLED=1`, so in every published release this feature is dead on macOS. It works only for people who `make build` themselves.

Erroring cleanly is the right behaviour for that build, so nothing here is a bug. But it means the headline feature effectively ships Linux-only, and turning cgo on for darwin isn't a config flip — cross-compiling from the Linux runner needs osxcross, or the release job needs a macOS runner.

Either shipping it as-is with a note in the README or adding a macOS build job seems fine to me. Just better as a decision than a surprise after release.

ianswope:
Those three are clean now, thanks. One more my list missed: `internal/player/local/player_darwin.go` — the `Player` struct fields at `:196` want realigning. Different directory, so the `gofmt -w internal/provider/local/` I suggested didn't reach it; `gofmt -w .` from the root covers it.

It was already dirty before your last two pushes, so nothing you introduced. It's the only file `gofmt` still flags at `9320d79`.

ianswope:
The mechanical side is clean at `5e98320`: `gofmt` reports nothing, `golangci-lint` v2.11.4 reports 0 issues, and `go build`, `go vet` and `go test ./...` all pass on linux/amd64. That closes out my last list.

@simonepelosi two things for you.

The macOS pipeline question is answered, just not visibly. Run `31504103360` (on `91a1d14`) did execute once approved, and **both `build-vet-test-macos` and `build-vet-test-linux` passed**. The only red job was `pre-commit`, on the gofmt items that are now fixed. The three runs since (`1d4d2d7`, `9320d79`, `5e98320`) are all sitting at `action_required` and will not report anything until they get an approve-and-run.

The `.goreleaser.yml` CGO question from my earlier comment is also still open, and it feeds into the next paragraph.

@GurKalra one thing my earlier review missed. I tested on macOS and wrote every finding against `player_darwin.go`, so that is where they all got fixed. Items 5, 6 and 7 apply just as much to `internal/player/local/player.go`, and that file still has all three. Driving the linux queue API directly:

```
SetQueue([C,B]) on queue [A B C]   -> queue=[A B C] len=3 idx=2   (want len=2)
clear + AppendQueue([A,B])         -> queue=[] len=0              (want len=2)
Next() on 1-track queue, RepeatOff -> wrapped to idx=0 and replayed
```

`AppendQueue` resolves incoming IDs against `p.queue` (`player.go:209`), so `:clear` still bricks playback until restart. `SetQueue` searches the existing queue and jumps to the first match instead of replacing it (`:167`). `Next()` wraps unconditionally and never consults `RepeatMode` (`:121`). The linux `Player` has no `allTracks` field at all.

The darwin version is the model for all three: `allTracks` populated by `LoadTracks`, `tracksForIDs(p.allTracks, ...)` in both `SetQueue` and `AppendQueue`, and the `RepeatModeOne` / `RepeatModeOff` branches at the top of `Next()`. Worth having parity, especially since the goreleaser point above means linux is the platform this feature actually ships on.

Two smaller ones:

- `player_darwin.go:421` still has a literal `44100`: `frames := C.SInt64(pos.Seconds() * 44100)`. Duration and position both derive from the real rate now, so seek is the last place the two disagree. `p.sampleRate` is already on the struct.
- The OGG scoping is right in code, but the claim is still unqualified in two places: the package doc at `internal/provider/local/provider.go:2` and `ROADMAP.md:21`.

Happy to re-run the full macOS matrix once there is a green run to compare against.

ianswope:
@GurKalra mechanical state at `9cd4c1d` on linux/amd64, since CI still has not run: `go build ./...`, `go vet ./...` and `go test ./...` all pass, `gofmt -l` is empty, `git diff --check` is clean, and `golangci-lint` v2.11.4 reports 0 issues. Nothing outstanding on the formatting side.

Two things in the queue port, both small.

**`Next()` does not stop the audio when it stops the queue.** `player.go:128` and `player_darwin.go:382` set `Playing = false` and broadcast, but never touch the pipeline. That is correct for the EOS path, where the stream has already ended. It is wrong for the manual one: `n` in the TUI (`internal/tui/model.go:2091`) and MPRIS `Next` (`internal/player/mpris/server.go:122`) both land in the same branch, so skipping past the last track with repeat off leaves the current track audible while the UI reports stopped, and `pollState` stops updating the position because it early-returns on `Playing == false`. On Linux this is new in this commit, since `Next()` used to wrap around instead. Adding `p.gst.Stop()` there, and `C.vibez_stop(p.audio)` on the darwin side, covers it.

**`AppendQueue` reads `p.allTracks` outside the lock** (`player.go:217`, `player_darwin.go:486`), while `SetQueue` does the same lookup inside it at `player.go:186`. Not reachable today: `LoadTracks` runs once at `cmd/root.go:120`, before `tea.NewProgram(...).Run()`, so nothing writes that field while the UI is live. It turns into a real race the moment a rescan or a music-directory change exists. Moving the `tracksForIDs` call below `p.mu.Lock()` is the whole fix.

The OGG scoping is consistent: `exts_linux.go` carries four extensions, `exts_darwin.go` three, `wantAudioFileCount` matches each, and `ROADMAP.md:21` and `provider.go:2` agree.

ianswope:
@simonepelosi the approval queue is what is holding this up now. Four runs are sitting at `action_required` with 0s duration: `1d4d2d7`, `9320d79`, `5e98320` and `9cd4c1d`. The last run that actually executed was `31504103360` on `91a1d14` on 11 August, where both `build-vet-test-linux` and `build-vet-test-macos` passed and only `pre-commit` was red, on gofmt items that are now fixed.

I ran the current head locally on linux/amd64: build, vet, test, `gofmt` and `golangci-lint` v2.11.4 are all clean. Approving the run on `9cd4c1d` would give you real signal for the first time in four pushes.

The `.goreleaser.yml:38` question from 11 August is also still open, and is the one thing here that is not GurKalra's to fix.

ianswope:
Re-ran the macOS matrix on `9cd4c1d`, Apple Silicon, go1.26.1 darwin/arm64 — and this time with **no local patches at all**. Both compile fixes from the August round are genuinely in: the scaffolding commit I needed on `b306c6a` is obsolete and I dropped it.

| Check | Result |
|---|---|
| `CGO_ENABLED=1 go build ./...` | clean (was `conflicting types for 'vibezOnEOS'`) |
| `CGO_ENABLED=0 go build ./...` | clean (was `LoadTracks`/`SetQueue`/`AppendQueue` undefined) |
| `go vet ./...` | clean |
| `go test ./...` | 17 packages ok |
| `gofmt -l` | empty |

**Five of the eight runtime defects from the August matrix are fixed**, verified on hardware rather than by reading:

| Item | Then (`b306c6a`) | Now (`9cd4c1d`) |
|---|---|---|
| Seek mid-playback | dead permanently | fixed — position advances 678ms → 2.786s |
| Duration at 48kHz | 5.44217687s for a 5.000s file | fixed — 5s, the hardcoded divisor is gone |
| Position accuracy | started at 557.278911ms, ran ~0.55s ahead | fixed — 1.478s at t=1.5s, ~22ms behind wall clock |
| `SetQueue` id list | ignored beyond `ids[0]` | fixed — `SetQueue([C,B])` → len=2, plays C |
| Play after `ClearQueue` | queue unrecoverable | fixed — Clear+Set+Append → len=3, playing |
| EOS auto-advance, mp3/flac/m4a, `#` in filenames | worked | still work |

Two known items are unchanged and still open: `.ogg` is scanned but unplayable (`duration=0s`, skipped after ~0.5–1.0s), and a single-track queue with repeat off still loops — `Next()` wraps via `% len(queue)` and `SetRepeat` is stored but never consulted.

## The one blocker: `Close()` during an auto-advance is a use-after-free

`playTrack` releases `p.mu` and *then* calls `C.vibez_start(audio)` on a local copy of the pointer (`internal/player/local/player_darwin.go:327`). `Close` takes `p.mu` and calls `C.vibez_destroy(p.audio)` on the same object (`:594`). The mutex guards the *field*, not the in-flight C call, so a `Close` landing in that window frees the AudioQueue/ExtAudioFile out from under a running `vibez_start`. One crash dump has both frames on the same address:

```
_Cfunc_vibez_start(0x81b548140)     ← goroutine 25, eosLoop → Next → playTrack
_Cfunc_vibez_destroy(0x81b548140)   ← the Close() on the test goroutine
SIGSEGV: segmentation violation ... signal arrived during cgo execution
```

Reproduction rates, polling every 5ms and calling `Close()` the instant the track change becomes observable (`pos=0s` in the state read is the marker for being inside the window):

| Corpus | Rate |
|---|---|
| `one.mp3` (5s) → `two.flac` (5s) | **4/4** |
| `short.mp3` (0.4s) → `two.flac` | 4/6 |
| `four.ogg` → `one.mp3` | 5/5 |

I chased the ogg angle first because that is where it surfaced, but **the ogg is not the cause** — it only phase-locks the collision with that probe's 750ms poll, which is why it looked deterministic there. Two controls separate a race from state corruption: the same advance with no `Close()` survives, and a fixed-delay `Close()` landing *after* the window survives 10/10 across 300/600/900/1500/3000ms.

In the app this is quit-during-auto-advance: press `q` as a track rolls over and the process can die inside CoreAudio instead of exiting.

I'm deliberately not proposing the patch — the invariant is that the C call must not outlive the lock that guards the pointer it uses, and how you satisfy that (hold `p.mu` across `vibez_start`, refcount the audio object, or have `Close` drain the EOS loop before destroying) is a design call that should stay yours.

## Observation, not a defect

With a 5ms poll, the change off `one.mp3` (reported duration 5.041632653s) is visible at t≈4.845s — roughly 197ms early. The probes run at volume 0, so whether the tail is actually truncated needs someone to listen. Worth a check while you're in here.

## Harness

Everything above is reproducible from `test/pr62-9cd4c1d` on my fork: the probes, the corpus generator, and the full then/now table in `TESTPLAN-MACOS.md`.

https://github.com/ianswope/vibez/tree/test/pr62-9cd4c1d

That branch is `9cd4c1d` plus test-only commits — no changes to your code. The new crash repro is `TestProbeCloseOnTrackChange` in `internal/player/local/zz_probe5_darwin_test.go`. All probes skip unless `PROBE_MUSIC_DIR` is set, and the corpus is generated by `scripts/gen-probe-corpus.sh` (six 5.000s ffmpeg sine files, one at 48kHz on purpose), so nothing binary is committed.

@simonepelosi one note while the run is queued: approving it gets you the mechanical checks, but macOS CI would not have caught this crash either way — the probes skip without an audio corpus, and there isn't one in CI. Hardware review stays necessary for this PR specifically.

@GurKalra the offer stands — push whenever you're ready and I'll re-run the matrix so you don't need a Mac.

GurKalra:
Fixed the Close()/playTrack() use-after-free with a sync.WaitGroup. eosLoop holds the WaitGroup count for its lifetime
- New() does audioWg.Add(1) before launching it, and eosLoop defers audioWg.Done(). 
- Close() calls audioWg.Wait() before destroying the audio object, so no Next() to playTrack() can be in flight when vibez_destroy fires. 

Happy for you to re-run the crash repro against this.

ianswope:
@GurKalra thanks. This does close the repro I filed, and the reason is worth stating: `eosLoop`'s only exit is the `<-p.done` case, so `Wait()` at `:595` cannot return while `eosLoop` is inside `playTrack`, and the `vibez_destroy` at `:598` can no longer land on a running `vibez_start`.

I have not put it on hardware yet, so treat the rest as a read rather than a test result. I am posting before the run because the WaitGroup covers one of four paths into the same free, and it seems better to give you all four at once.

## `audioWg` counts only `eosLoop`

`playTrack` is reached from five other call sites in four methods, each of which drops `p.mu` and then calls `C.vibez_start(audio)` at `:330`:

| Caller | Line |
|---|---|
| `Next`, RepeatModeOne | `:380` |
| `Next` | `:395` |
| `Previous` | `:417` |
| `SetQueue` | `:467` |
| `SetPlaylist` | `:484` |

Those run on whatever goroutine the caller is on, and the TUI makes that a different goroutine from `Close` by construction. `n` dispatches `Next` through `playerCmd` (`internal/tui/model.go:2095`), which bubbletea runs as a `tea.Cmd` off the event loop, while `Close` is called synchronously inside `handleKey` (`:1162`) and `executeCommand` (`:1426`) on the event loop itself. Press `n`, then `q` before the change settles, and it is the original crash with `eosLoop` not involved.

## `ClearQueue` is a second destroyer

`:558-559` calls `vibez_stop` then `vibez_destroy` under `p.mu`, and the WaitGroup does nothing for it. Clearing the queue during an auto-advance frees the object `eosLoop`'s `playTrack` is about to start.

## `playTrack` races itself, with no `Close` involved

Each call keeps its own local `audio` and starts it after unlocking, while `:311` destroys `p.audio` under the lock. Two concurrent calls interleave as: A sets `p.audio = audioA` and unlocks, B destroys `audioA` and installs `audioB`, then A calls `vibez_start(audioA)` on freed memory. The `p.handle.Delete()` and `cgo.NewHandle(p)` pair at `:316-318` has the same problem.

## Invariant

Unchanged from my last comment: the C call must not outlive the lock that guards the pointer it uses. A WaitGroup satisfies that for the one goroutine it counts. Holding `p.mu` across `vibez_start`, or refcounting the audio object, satisfies it for all of them. Which of those you pick is still your call.

One cosmetic thing, not a bug: `select` picks randomly when both `eosCh` and `done` are ready, so `Close` can still let one more track start and then immediately destroy it. Wasted work, nothing more.

## Linux, since 9cd4c1d ported the queue fixes there

`gst.Destroy()` (`internal/player/gst/player.go:202`) is already the same shape as your new code: `close(stop)`, `wg.Wait()`, then destroy under the lock, with `p.wg` counting only `busLoop`. `p.pipeline` is not cleared after `vibez_gst_destroy`, so a caller-facing method that takes the lock after `Destroy` gets a freed pipeline. I have not tried to reproduce it, so treat that as worth a look rather than a report.

## Harness

The probe branch is rebased onto `a34eae3` and carries a repro for each of the three windows above, in `internal/player/local/zz_probe6_darwin_test.go`:

https://github.com/ianswope/vibez/tree/test/pr62-a34eae3

| Probe | Window |
|---|---|
| `TestProbeCloseDuringManualNext` | user `Next()` vs `Close()` |
| `TestProbeClearQueueOnTrackChange` | EOS advance vs `ClearQueue()` |
| `TestProbeConcurrentPlayTrack` | `playTrack` vs `playTrack`, no `Close` |

Each takes the process down when it hits, so they run one at a time. `PROBE_NEXT_CLOSE_DELAY_MS` and `PROBE_CLEAR_DELAY_MS` push the call past the post-unlock window, which is the control that separates a race from state corruption.

One correction to what I posted earlier: the harness was not quite self-contained. `gen-probe-corpus.sh` wrote six 5.000s files and no `short.mp3`, which `TestProbeCloseAtEOSStart` and `TestProbeCloseOnTrackChange` both default to, so on a freshly generated corpus those two opened a missing file and reported "survived" without reaching the window. The generator now writes it at 0.4s. If you pulled the corpus script before today, regenerate.

I will run all of it on `a34eae3` next time I am booted into macOS and post the numbers.

simonepelosi:
@GurKalra @ianswope Sorry for the late feedback, I had a really really hard week, I'm fully back now

ianswope:
@GurKalra ran the matrix on `a34eae3`, Apple Silicon, go1.26.1 darwin/arm64, no local
patches. Your WaitGroup fix holds, and this round that is reproduced rather than read.
`TestProbeCloseOnTrackChange` is the repro I filed last round, and I re-ran all three of
the track configurations it went down on:

| First -> second | On `9cd4c1d` | On `a34eae3` |
|---|---|---|
| `one.mp3` -> `two.flac` | crashed 4/4 | **survived 6/6** |
| `short.mp3` -> `two.flac` | crashed 4/6 | **survived 6/6** |
| `one.mp3` -> `four.ogg` | crashed 5/5 | **survived 6/6** |

Eighteen runs, every one confirmed to have reached the track change, no crash in any of
them. `eosLoop`'s only exit being the `<-p.done` case is what makes that hold: `Wait()`
cannot return while `eosLoop` is inside `playTrack`.

The three windows I flagged from reading also reproduce.

## Mechanical state

| Check | Result |
|---|---|
| `CGO_ENABLED=1 go build ./...` | clean |
| `CGO_ENABLED=0 go build ./...` | clean |
| `go vet ./...` | clean |
| `gofmt -l .` | empty |
| `go test ./...` | 17 packages ok |

Nothing outstanding on the mechanical side, and CI is green on this head now.

## The three windows `audioWg` does not cover

Every one of these is the same signature: `SIGSEGV`, "signal arrived during cgo
execution", with `_Cfunc_vibez_start(P)` and `_Cfunc_vibez_destroy(P)` on two threads at an
identical `P`.

| Window | Probe | Crashed | Control | Control result |
|---|---|---|---|---|
| user `Next()` vs `Close()` | `TestProbeCloseDuringManualNext` | **14/15** | `Close` +500ms | survived 10/10 |
| EOS advance vs `ClearQueue()` | `TestProbeClearQueueOnTrackChange` | **7/7** | `ClearQueue` +500ms | survived 10/10 |
| `playTrack` vs itself, no `Close` | `TestProbeConcurrentPlayTrack` | **7/7** | 1 goroutine, same call count | survived 10/10 |

Each control moves the second call clear of `playTrack`'s post-unlock window, and the
crash stops, which separates a race on the pointer from ordinary state corruption. Window 1's single survivor out of fifteen was
a timing miss; the log shows it armed correctly (`track="B" playing=true`) and simply lost
the race that run.

So `audioWg` closes the `eosLoop` path and only that path. `Close` still races a
user-initiated `Next`, `ClearQueue` is still a second destroyer the WaitGroup never
consults, and two concurrent `playTrack` calls still free each other's object with no
`Close` anywhere in the picture.

## Invariant

Unchanged, and still your call: the C call must not outlive the lock that guards the
pointer it uses. A WaitGroup satisfies that for the one goroutine it counts. Holding
`p.mu` across `vibez_start`, or refcounting the audio object, satisfies it for all of
them.

## Correction to my own harness

If you ran the round-3 snippet I posted, throw the results away and pull again. I had
`export PROBE_MUSIC_DIR=./probe-corpus`, and it has to be absolute: `go test` runs each
test binary in its own package directory, so the relative path resolved against
`internal/player/local/` instead of the repo root. Track IDs pointed at nothing,
`LoadTracks` produced an empty queue, `SetPlaylist` returned `nil` at its
`len(p.queue)==0` guard, and every `title != "A"` guard then fired at t≈6ms against
`track="" playing=false`. All three probes printed "survived" without ever reaching the
window. My own first pass "passed" 4/4 that way before I looked at the state in the log.

`musicDir` now resolves the path and refuses one it cannot read, and `mustPlay` fails any
probe whose playback never started, which catches this and the missing-`short.mp3` bug
from last round in one place. Rounds 1 and 2 used an absolute `/tmp/probe-corpus`, so
those published numbers stand.

Second correction: I described `PROBE_PLAYTRACK_ROUNDS` as window 3's control. It is not.
Every round launches two concurrent `Next()` calls, so `ROUNDS=1` still races two of them
and still dies, 10/10. The real knob is `PROBE_PLAYTRACK_GOROUTINES=1`, same total call
count with no concurrency, and that is the 10/10 survival in the table.

## Early EOS, sharper than last round

I flagged this as unconfirmed before. It reproduces consistently, and the number itself
is the finding:

| Track | Duration | Advance seen at | Early by |
|---|---|---|---|
| `one.mp3` | 5.041632653s | 4.854s | ~188ms |
| `short.mp3` | 0.400000s | 0.204s | ~196ms |

Roughly 190ms early regardless of track length, across a 12x difference in duration. A
proportional error would scale; a constant one points at a fixed prefill or buffer count
being dropped from the end rather than a rounding bug in the duration maths. Still needs
a human ear to say whether the tail is actually clipped, since the probes run at volume 0.

## Harness

Rebased on `a34eae3` with the fixes above:
https://github.com/ianswope/vibez/tree/test/pr62-a34eae3

```sh
./scripts/gen-probe-corpus.sh ./probe-corpus
export PROBE_MUSIC_DIR="$PWD/probe-corpus"
go test ./internal/player/local/ -run TestProbeCloseDuringManualNext   -v -count=1
go test ./internal/player/local/ -run TestProbeClearQueueOnTrackChange -v -count=1
go test ./internal/player/local/ -run TestProbeConcurrentPlayTrack     -v -count=1
```

One at a time; the fault takes the process down, so a shared run cannot tell you which
probe died. I can re-run this on any future push, so you do not need a Mac for it.

The Linux `gst.Destroy()` note from my last comment is unchanged and still unverified: it
is the same shape as your new code, and `p.pipeline` is not cleared after
`vibez_gst_destroy`. Worth a look rather than a report.

simonepelosi:
Hey @GurKalra !

#120 (thanks @ianswope) is now merged into main, so the macOS release runner with CGO_ENABLED=1 is officially in place.

Regarding the latest push for the audioRef refcounting: macOS CI (build-vet-test-macos) failed due to 4 leftover audio variable references in internal/player/local/player_darwin.go:

Line 270: C.vibez_get_playback_time(audio) $\rightarrow$ ref.ptr
Line 336: C.vibez_get_sample_rate(audio) $\rightarrow$ raw
Line 341: C.vibez_get_duration(audio) $\rightarrow$ raw
Line 370: C.vibez_start(audio) $\rightarrow$ ref.ptr
Once those are updated and the branch is rebased on main, CI should be green across both Linux and macOS and we should be ready to have the final review!

ianswope:
Rebased the probe harness onto `9fe56b4` (`test/pr62-9fe56b4` on my fork) and type-checked it against this head on Linux. The probes need no changes; they use the public API plus `p.mu`, `p.queue` and `p.idx`, which are all still there.

One thing to fix before anyone runs them, or every probe dies on its first state read and looks like a use-after-free reproduction:

`player_darwin.go:638-640`

```go
p.mu.RLock()
s := p.state
p.mu.Unlock()
```

`2c665fd` changed the `Lock` to `RLock` and left the `Unlock`. `p.mu` is a `sync.RWMutex`, and `Unlock` after `RLock` is `fatal error: sync: Unlock of unlocked RWMutex`, a runtime throw rather than a panic, so no `recover` catches it. Nothing in the app calls `GetState` today, so the TUI will not hit it, but it is on the `Player` interface and the first caller takes the process down. The probes call it constantly. `go vet` cannot see it, there are no tests for this package, and CI has not run on this head, so nothing was going to flag it. It is the only `RLock`/`Unlock` mismatch in the file; the Linux `player.go` is unaffected.

The line numbers in the probe comments and TESTPLAN drifted with the `audioRef` refactor (`vibez_start` is now `:370`, and `Close` and `ClearQueue` no longer call `vibez_destroy` directly). Comments only. I will bring them up to date on the next harness round, after the fix above lands.

ianswope:
Ran the probe harness against `74319b9` on Apple Silicon, macOS 26.3, go1.26.1 darwin/arm64, Apple clang 21.0.0.

No local patches this time. Both `CGO_ENABLED=1 go build ./...` and `CGO_ENABLED=0 go build ./...` succeed on the head as it stands, so the compile scaffolding I have been carrying since August is no longer needed and I have dropped it. `go vet` is clean on both local packages, and `go test ./internal/provider/local/` passes.

All 18 probes ran in separate processes with `-count=1`. Every one passed and nothing panicked.

Five things that were broken in August are fixed:

| | August (`b306c6a`) | `74319b9` |
|---|---|---|
| Duration at 48kHz | 5.44217687s, the hardcoded 44100 divisor | 5s |
| Position accuracy | ~0.55s ahead, reporting the decode cursor | 1.477s at t=1.5s |
| Seek mid-playback | killed playback permanently | recovers, position keeps advancing |
| `SetQueue` id list | ignored past `ids[0]` | `SetQueue([C,B])` gives len=2 playing C |
| Play after `ClearQueue` | queue unrecoverable without a restart | rebuilds and plays |

The refcounting work holds up under everything I could throw at it: 300 rapid transitions, 6 deliberate EOS-boundary collisions, 40 rounds of two goroutines in `playTrack`, and `Close()` fired at EOS start, on a track change, and during a manual `Next`, plus `ClearQueue` on a track change. No panic in any of them.

Two things still open.

**Repeat-off still loops.** With `RepeatMode=Off` and a single track queued, it restarted at 0.8s and carried on. `Next()` wraps via `% len(queue)` and `SetRepeat` is stored but never read.

**`.ogg` reports `duration=0s`** and gets advanced past in under a second, same as August. There is no Vorbis decoder on the CoreAudio path, so this may be a "don't scan it" decision rather than a playback one. Your call.

One smaller observation. After EOS on the last track, `Seek(1s)` does resume audio and the position advances, but `Playing` stays `false` the whole time. August had the inverse, position frozen with `Playing` stuck true, so the state flag and the audio still disagree in that window, just in the other direction now.

simonepelosi:
Thanks @GurKalra! The atomic `audioRef` refcounting in `player_darwin.go` and the recent `RUnlock` fix make the CoreAudio backend much more solid under concurrency and stress testing.

A few technical items to address before merging:

1. **AudioQueue buffer leak on seek (`player_darwin.go`)**:
   `vibez_seek` calls `AudioQueueAllocateBuffer` in a loop for all 3 buffers on every seek without freeing the existing buffers allocated during `vibez_start` or previous seeks. In CoreAudio, `AudioQueueAllocateBuffer` memory is only released on `AudioQueueDispose` or explicit `AudioQueueFreeBuffer`. Reusing an array of preallocated buffers in `vibez_audio_state` (or calling `AudioQueueFreeBuffer`) will prevent memory growth on repeated seeks.

2. **Seek unpauses unconditionally (`player_darwin.go`)**:
   `vibez_seek` currently calls `AudioQueueStart(s->queue, NULL)` unconditionally if `!s->done`. If playback was paused or stopped after EOS, seeking will resume audio output in CoreAudio while Go's `p.state.Playing` remains `false`. `vibez_seek` should only start the queue if the player was actively playing.

3. **`.ogg` on macOS (`internal/provider/local/exts_darwin.go`)**:
   CoreAudio's `ExtAudioFile` doesn't ship with native Ogg Vorbis decoding. `.ogg` files in scanned directories report `duration=0s` and immediately skip. We should remove `.ogg` from `exts_darwin.go` (keeping it in `exts_linux.go` where GStreamer supports it).

4. **`Makefile` `check-deps` breaks on macOS**:
   `check-deps` checks for `gstreamer-1.0`, `gtk+-3.0`, and `webkit2gtk-4.1`. Since macOS uses CoreAudio and Chrome/CDP, running `make build` on macOS fails on these missing Linux pkg-config packages. `check-deps` should only run on Linux.

5. **`time.Sleep(200ms)` in Linux `playTrack` (`internal/player/local/player.go`)**:
   Sleeping synchronously to query GStreamer duration blocks callers and can fail on slow I/O. It's better to let `pollState()` update duration once GStreamer reaches `PLAYING`.

6. **Rebase**:
   Please rebase on `main` to catch up with recent releases.

Once these are addressed, we should be good to merge!

simonepelosi:
Thanks for the quick turnaround on commit `8958ea9` @GurKalra! Reusing preallocated buffers in `vibez_seek`, guarding `check-deps` for Linux, removing the sleep in `playTrack`, and the `RUnlock` fix make this much cleaner.

Just two small items left before we can merge:

1. **Reset `s->playing` on EOS (`player_darwin.go`)**:
   In `vibez_audio_callback`, when `vibez_fill_buffer` returns EOF, `AudioQueueStop` is called and `s->done = 1`, but `s->playing` remains `1`. If playback stops at the end of the queue and the user later seeks, `vibez_seek` checks `if (!s->done && s->playing)` which restarts CoreAudio output while Go's `p.state.Playing` is `false`. Setting `s->playing = 0;` in `vibez_audio_callback` when `vibez_fill_buffer` hits EOF resolves this.

2. **Drop auto-play on launch (`cmd/root.go`)**:
   Lines 123-129 in `cmd/root.go` currently call `plyr.SetQueue(ids)` to auto-play track 0 immediately upon opening the app. In standard and demo modes, vibez opens with the library loaded and waits for user interaction before playing audio. Removing `plyr.SetQueue(ids)` leaves the tracks loaded in the TUI queue via `opts.InitialTracks`, allowing the user to press Enter/Space when they want to play.

3. **Rebase**:
   Please rebase on `main` to pick up recent fixes.

Once these are updated, happy to merge!

simonepelosi:
Thanks a lot @GurKalra! Both fixes look great and the clean rebase on `main` is much appreciated. Everything is passing CI and looks solid to me.

@ianswope does this look good to you as well for a final sanity check before we merge?

ianswope:
Ran the probe harness against `8b5b18b` on Apple Silicon, macOS 26.3, go1.26.1 darwin/arm64. All 18 probes in separate processes with `-count=1`. `CGO_ENABLED=1` and `CGO_ENABLED=0` both build, `go vet` clean, `go test ./...` 19 packages ok and 0 failures. Harness is on `test/pr62-8b5b18b` on my fork: the probe files on top of this head, nothing else touched.

**First, a caveat on how to read these probes.** They are observational, not assertive: several log whichever outcome they see and pass either way. So "18/18 pass" is not on its own evidence that anything was fixed, and my run on `74319b9` last week was 18/18 while this EOS bug was still present. The measurements below are the signal.

**EOS resetting `s->playing`.** Confirmed by A/B against the parent commit, `90680db`:

| head | after `Seek(1s)` + 1.5s wait | |
|---|---|---|
| `90680db` (before) | `pos=1.480090702s playing=false` | audio resumed behind Go's back |
| `8b5b18b` (after) | `pos=1s playing=false` | holds, no restart |

That is the bug and its fix. Note the probe's own log line on the fixed head reads `playback appears dead after EOS`, stale wording from when this was a suspected defect, and now the behaviour you asked for.

**The 48kHz divisor.**

| file | duration | position at 1.5s |
|---|---|---|
| one.mp3 (44.1kHz) | 5.041632653s | 1.477505668s |
| two.flac (44.1kHz) | 5s | 1.478072562s |
| three.m4a (44.1kHz) | 5s | 1.477981859s |
| six_48k.flac (48kHz) | 5s | 1.478299319s |
| four.ogg | 0s | 457.210884ms |

`six_48k.flac` at 5s is the point; a hardcoded 44100 divisor gives 5.44s.

**`.ogg` on darwin.** Removed from `exts_darwin.go`, kept in `exts_linux.go`. The probes bypass the scanner and feed paths directly, so they still exercise the old path and show it degrades rather than hangs: `TestProbeOggNeverEnds` advances off after 0.8s, and the table above shows why it had to go.

Worth knowing rather than fixing here: the scan skips unsupported extensions silently (`provider.go:51`), so a macOS user with an ogg library sees those files simply not appear, with no message. Still better than listing them and skipping instantly.

**Buffer reuse on seek.** `TestProbeSeekMidPlayback` and `TestProbeHandleChurn` pass; `TestProbeEOSRaceChurn` survived 6 EOS-boundary collisions without a panic.

**Auto-play on launch.** Gone from `cmd/root.go`, `InitialTracks` still populates the TUI queue.

Good to merge from my side.

One practical note: I am travelling for the next day or so, so I may be slow coming back on anything that needs a follow-up run here. The results above stand on their own, so please do not hold the merge for me.

simonepelosi:
Thank you so much @GurKalra for this major feature and your great work through all the iterations! Huge thanks as well to @ianswope for the extensive real-hardware testing and probe harness.

Merging now! 🎉

PR #77, opened 2026-07-07
fix: emit MPRIS Seeked signal so clients resync after a seek

### LINKED ISSUE #73: Bug: MPRIS Integration shows incorrect track position
When opening the media player widget in KDE, the track position is wrong - it only counts up when the widget is open.

For example, when I start a track and let it run for 30 seconds, and then open the media player, it reports that the track is at 0 seconds, and then starts counting.

In the media player:
Image

In `vibez`:
Image

Since this works correctly with other apps (e.g. Sidra, Chrome) I'm assuming this is a bug in `vibez`.

## Environment

OS: Arch Linux 7.0.11
DE: KDE Plasma 6.7.1
WM: KWin (Wayland)
App version: `latest` (`vibez v0.3.0`)

Soejii:
vibez's MPRIS server never emits `org.mpris.MediaPlayer2.Player.Seeked` and only writes `Position` on track/status changes. Per the MPRIS spec, clients resync position only on `Seeked`, so seeking within a track leaves external clients drifting on a stale position until the next track.

`flush()` now detects a position jump within the current track and emits `Seeked` with the new position. The check is a small pure helper (`isSeek`) with unit tests, and the 2s threshold sits between the ~1s update cadence and vibez's ±10s seek step.

I kept this inside the MPRIS server by inferring the seek from a position discontinuity. If you'd rather have the engine signal seeks explicitly, I'm glad to rework it.

Verified locally (cgo build with the repo's pkg-config shim): `go build`, `go vet`, and `go test ./...` all pass including the new tests; gofmt clean, `go mod tidy` a no-op.


simonepelosi:
Thank you so much for your contribution, I think it's ready to be merged! Can you also rebase on main and update the changelog with this fix ? @Soejii

Soejii:
Done! Rebased on the latest main and added a changelog entry under Unreleased → Fixed. Thanks for the review

simonepelosi:
Thank you so much for your first contribution! I'm so happy and glad for that. @Soejii

Issue #81, opened 2026-07-07
v0.4.0: 401 when adding a playlist to the queue (works fine on v0.3.1)


ianswope:
First off, thanks for vibez — really nice to have a proper Apple Music TUI on Linux.

Hit a bug on v0.4.0 (fresh install from the install script earlier today). Search works fine, but the moment I add a playlist to the queue I get:

`search: apple music api 401 Unauthorized. Re-run vibez login...`

How I hit it:
- `vibez auth login` — works, `vibez auth status` shows "Authenticated with Apple Music (storefront: us)"
- search "Smithereens", results load fine
- pick a playlist from the results, add it to the queue → 401

I'm pretty sure it's not actually an auth problem, though. I pulled the developer and user tokens straight out of `~/.config/vibez/config.json` and hit the Apple API myself with curl — everything comes back 200:
- catalog search, catalog playlist, and catalog playlist tracks (developer token)
- `/v1/me/library/playlists` and `/v1/me/storefront` (developer + user token)

The developer token JWT isn't expired either — about 30 days left on it. So Apple's happy with the credentials; something about how 0.4.0 builds the request while resolving the queue seems to be the issue.

I rolled back to v0.3.1 and add-to-queue works fine with the same account and the same playlist, so this looks like a 0.4.0 regression. My guess is it came in with the large-playlist fix from #74 — the new parallel batch track resolution / pagination path in the 0.4.0 notes is exactly what add-to-queue hits, so it seems like that rewrite fixed the "large playlists don't load" case but broke queue resolution with a 401.

Happy to grab a `--debug` log if you point me at where it writes during a TUI session.

No rush on any of this, 0.3.1 is working great for me in the meantime. Thanks again for building it, really glad it exists.

simonepelosi:
Hey! Thank you for reporting that, sorry I closed it by mistake but the fix should be easy this time, I released the new 0.4.1! Can you please update your version and let me know ? <3 @ianswope

ianswope:
Confirmed fixed on v0.4.1 — adding a catalog playlist to the queue works perfectly now, no more 401. Thanks for the quick turnaround! ❤️

simonepelosi:
You're very welcome! ❤️

PR #51, opened 2026-05-22
Add ASCII album art to now-playing panel

elijah-rou:
## Motivation
Closes #33. The now-playing panel can show album artwork as terminal-native block art when truecolor is available.

## Summary
- Add pure Go PNG/JPEG artwork decoding and 24-bit ANSI half-block rendering
- Fetch artwork asynchronously, cache decoded images by URL, and cache rendered sizes for terminal resizes
- Render artwork left of now-playing metadata/progress with text-only fallback when truecolor or artwork fetch is unavailable
- Add demo artwork URLs so demo mode exercises the same artwork download/render path without bundling album-art files
- Add unit coverage for exact renderer output, truecolor fallback, stale artwork guards, fetch failures, and demo artwork metadata

## Tests
- go test ./...
- CI: build-vet-test-linux, build-vet-test-macos, pre-commit

## Examples

screenshot-2026-05-22_12-32-05
screenshot-2026-05-22_12-32-17

PR #83, opened 2026-07-08
feat(tui): render album art in the now-playing panel (closes #33)

### LINKED ISSUE #33: feat: ASCII album art in the now-playing panel

ianswope:
## What

Implements the album art request from #33: the current track's cover now renders as coloured Unicode half-block art on the left of the **Now Playing** panel, with the track info centred to its right.

Each character cell is an upper-half block `▀` whose **foreground is the top pixel** and **background is the bottom pixel**, so one cell encodes two vertically-stacked image pixels. Because a terminal cell is ~twice as tall as it is wide, this yields roughly square pixels — a square cover reads as square when drawn with `cols == rows*2`.

## Why

The now-playing panel was text-only; a cover makes it feel like a real player. This is the enhancement filed in #33.

## How

- **`internal/tui/art`** — a small, pure-Go converter (`Decode` + `Render`) using only stdlib `image/jpeg` + `image/png`. Colours are emitted through `lipgloss`, so the active renderer down-samples them to the terminal's colour profile automatically.
- The cover is fetched **asynchronously** (a `tea.Cmd`) whenever the artwork URL changes; the `ArtworkURL` already present on `player.State.Track` is reused, so no provider changes were needed.
- Rendered half-block lines are **cached** (keyed by URL + dimensions) so they only re-render on a track change or a resize — not every frame.
- Layout **adapts to width** and falls back to the existing centred layout when the panel is too narrow. Text lines are ANSI-aware truncated so a long title can't overflow and break the box border.
- Art (and its download) are **skipped entirely** on terminals without at least 256-colour support, per the graceful-fallback note in the issue.
- New config option **`album_art`** (default `true`) to turn it off.

The no-art code path is unchanged (the width variable simply stays the full content width), so existing behaviour is byte-identical when art isn't shown.

## Tests

- `internal/tui/art`: decode round-trip, invalid input, output dimensions/width, colour mapping (top→fg, bottom→bg), and guard cases.
- `internal/tui`: `nowPlayingLines` renders art on the left with track text on the right, respects the config toggle, skips art without truecolor / when too narrow, and ignores a stale cover whose URL doesn't match the current track.
- Verified the full fetch → decode → render pipeline against a **real Apple Music cover** end-to-end and confirmed the reconstructed output is a faithful cover.
- `go build ./...`, `go vet ./...`, `go test ./...`, and `golangci-lint` (v2.11.4, matching CI) all pass clean.

Happy to adjust the default cover size, the colour-profile threshold, or the config surface to taste.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,8 @@ require (
 	charm.land/bubbles/v2 v2.1.0
 	charm.land/bubbletea/v2 v2.0.6
 	charm.land/lipgloss/v2 v2.0.3
+	github.com/charmbracelet/colorprofile v0.4.3
+	github.com/charmbracelet/x/ansi v0.11.7
 	github.com/godbus/dbus/v5 v5.2.2
 	github.com/golang-jwt/jwt/v5 v5.3.1
 	github.com/playwright-community/playwright-go v0.6000.0
@@ -15,9 +17,7 @@ require (
 
 require (
 	github.com/atotto/clipboard v0.1.4 // indirect
-	github.com/charmbracelet/colorprofile v0.4.3 // indirect
 	github.com/charmbracelet/ultraviolet v0.0.0-20260416155717-489999b90468 // indirect
-	github.com/charmbracelet/x/ansi v0.11.7 // indirect
 	github.com/charmbracelet/x/term v0.2.2 // indirect
 	github.com/charmbracelet/x/termios v0.1.1 // indirect
 	github.com/charmbracelet/x/windows v0.2.2 // indirect
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/playwright-community/playwright-go v0.6000.0
 	github.com/spf13/cobra v1.10.2
 	github.com/webview/webview_go v0.0.0-20240831120633-6173450d4dd6
+	golang.org/x/sys v0.43.0
 )
 
 require (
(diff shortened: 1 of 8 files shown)


ianswope:
Tested this locally against my own Apple Music library (built with `make build-with-token`) and it's working great — covers render cleanly to the left of Now Playing, and the plain centred layout is untouched when art is off or the panel's too narrow. Happy to adjust the default cover size, colour-profile threshold, or anything else you'd like. Thanks again for the fast turnaround on #81 the other day! 🙏

ianswope:
Closing in favour of #51, which predates this and is already in discussion — apologies, I opened this without spotting #51 first.

@elijah-rou if useful, two bits from my attempt you're welcome to: (1) area-averaging the downscale instead of nearest-neighbour (noticeably less pixelated at small sizes), and (2) sizing the cover to the terminal's measured cell aspect ratio via TIOCGWINSZ so it renders a true square on wide fonts like JetBrains Mono. Happy to help however's useful. 🙏

PR #86, opened 2026-07-09
feat: add radio mode seeded from any track

### LINKED ISSUE #32: feat: radio mode — start a station from any track
## Summary

Press `r` on any track in the queue or search results to start an Apple Music radio station seeded by that track.

## Behaviour

- `r` on a focused track → calls `GET /v1/me/stations/next-tracks` (or the catalog radio endpoint) to fetch a list of related tracks
- Loads them into the queue and starts playback
- TUI shows a `📻 Radio` badge somewhere in the now-playing area to indicate radio mode is active
- Pressing `r` again (or manually editing the queue) exits radio mode
- Auto-refills the queue when it gets low (fetch next batch from the station)

## API

Apple Music API: `POST /v1/me/stations/next-tracks/{stationId}` or seed via `stationSeed` (type `songs`, id = catalog ID of the current track).

## Why it's cool

One keypress turns any song into an infinite, taste-matched radio stream without leaving the TUI.

ianswope:
## Summary
Closes #32.

- Press `R` on the currently playing track, a selected queue track, or a selected search result to start an Apple Music radio station seeded from it.
- Uses Apple's synthetic per-song radio station ID: `POST /v1/me/stations/next-tracks/ra.<catalogSongID>` (verified against the live API — the `stationSeed` body shape originally discussed in the issue turned out not to match Apple's actual endpoint; happy to elaborate on how I found the right one).
- Related tracks are appended to the queue and playback continues into them; the queue auto-refills as it runs low, reusing discovery mode's auto-refill/circuit-breaker machinery per our discussion in the issue.
- A `📻 radio` badge shows in the now-playing status bar while active, mirroring the existing discovery indicator.
- Radio and discovery are mutually exclusive (both compete for the same last-track refill trigger) — starting one stops the other.
- Search uses `ctrl+r` rather than a bare `R`, matching the existing `ctrl+p` precedent, since the search box is a live text input and a bare letter would steal from typing.

## Test plan
- [x] `go test ./...` and pinned `golangci-lint` (v2.11.4) both clean
- [x] Unit tests for `GetStationTracks` (Apple provider) and the model-layer keybindings/refill/retry/dedup logic
- [x] Manually verified in `--demo` mode: badge, queue refill, queue-panel seeding, `ctrl+r` from search, and that `r` (repeat) is unaffected
- [x] Manually verified against my real Apple Music account: played an album, pressed `R`, got real related tracks appended and playing, confirmed via debug log (`[radio] refilled N tracks`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - **Radio mode had no effect mid-album/playlist** — enabling radio while an album or playlist was already queued did nothing until the very last queued track played, since refills only triggered once the current track became the last item in the queue. Radio now drops the rest of the queued album/playlist as soon as it starts, so its picks actually play next.
+- **Dropped album/playlist tracks came right back on refill** — the tracks dropped when radio started were only removed from the queue, not blacklisted, so the very next refill's dedup (which only checks what's still queued) let the station API hand them straight back — since station results are often from the same album/playlist context as the seed. Radio now blacklists dropped tracks so they can't be re-added.
 
 ## [0.4.1] — 2026-07-08
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **Radio mode (`R`)** — press `R` on the currently playing track, a selected queue track, or a selected search result to start an Apple Music radio station seeded from it. Related tracks are appended to the queue and playback continues into them; the queue auto-refills as it runs low, and a `📻 radio` badge shows in the now-playing status bar while active. Press `R` again to stop. Closes #32.
+
 ## [0.4.1] — 2026-07-08
 
 ### Fixed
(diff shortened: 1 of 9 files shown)


simonepelosi (commented):
Hey, left some feedback during the review:

1. `m.radio.skipped` never gets cleared. So any tracks dropped during a radio session will stay blacklisted forever across all subsequent radio sessions in that run. We should probably reset/init `m.radio.skipped` in `startRadioFrom` and set it to `nil` in `stopRadio`.

2. When starting a radio from a search result (`ctrl+r`), the seed track itself is never added to the queue because `dropQueueAfter` returns early since the track isn't in the queue yet. So the seed is never played and we just append the radio picks to the end of the existing queue. We should probably insert the seed track as play next if it's not already in the queue.

Other than that, the tests look great and everything else is solid!

simonepelosi:
Can you also update the commands in the README ?

ianswope:
Fixed both:

1. `m.radio.skipped` now resets to `nil` in both `startRadioFrom` and `stopRadio`, so it can't carry stale entries into a later radio session.
2. `dropQueueAfter` now inserts the seed as play-next (reusing `playNextCmd`) when it isn't already in the queue, so radio started from a search result actually plays the seed instead of skipping straight to the picks.

Added regression tests for both. Also updated the README — radio mode and its key bindings (`R` global/queue, `ctrl+r` search) were undocumented.

simonepelosi:
It looks great! Please, if you can solve the conflicts and then I will merge it ❤️ thank you so much for that contribution

ianswope:
Resolved the merge conflicts — should be clear to merge now! 🎉

simonepelosi:
Thank you so much for your contribution ❤️

PR #91, opened 2026-07-14
build: add 'make refresh-token' to regenerate the dev token into config

ianswope:
Local (non-embedded) builds read `apple_developer_token` from `~/.config/vibez/config.json`, and a signed Apple MusicKit token expires after ~30 days — so a dev token that worked last month silently starts returning `invalid token` at launch. This adds a quick way to refresh it:

- **`gen-devtoken -write`** — saves the freshly signed JWT straight into the vibez config (optional `-config` path) instead of printing it to stdout.
- **`make refresh-token`** — signs a token from credentials in the environment or a gitignored `.env` (`APPLE_KEY_ID`, `APPLE_TEAM_ID`, and `APPLE_PRIVATE_KEY` or `APPLE_PRIVATE_KEY_FILE`) and writes it in.

Default stdout behaviour is unchanged, so `build-with-token` / `release` still work exactly as before. Adds a test for the `-write` path.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 GOFLAGS=-tags webkit2gtk_4_1
 PKG_CONFIG_PATH=$(CURDIR)/pkg-config
 
-.PHONY: build run test lint clean release build-with-token
+.PHONY: build run test lint clean release build-with-token refresh-token
 
 build:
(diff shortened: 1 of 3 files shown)


simonepelosi:
Thank you so much for helping with this <3

simonepelosi (approved):

Issue #11, opened 2026-04-15
Arm64 support


simonepelosi:
Support arm64 architectures

simonepelosi:
Closing as completed. Full-track playback on ARM64 Linux landed in #94, and Apple Silicon support landed in #116.

PR #94, opened 2026-07-17
Add Linux/arm64 full-track playback support

ianswope:
## Summary

Addresses #11. Google publishes no Chrome build for Linux/arm64, so the Chrome/CDP full-track backend was effectively amd64-only and arm64 fell back to WebKit 30-second previews. This adds an arm64 path that uses a system-installed Chromium plus a system-registered Widevine CDM instead of downloading a browser.

amd64 and macOS behavior is unchanged.

## How it works on arm64

- **Discovery instead of download.** On arm64, `EnsureBrowser` no longer downloads Chrome. It discovers a system browser (preferring the real `/usr/lib/chromium/chromium` binary over launcher wrappers, with `VIBEZ_CHROME_PATH` / `CHROME_PATH` overrides) and a registered Widevine CDM.
- **`cdp.Available()`** replaces the `runtime.GOARCH == "amd64"` gate in `root_linux.go`. CDP is selected on arm64 only when both a browser and a CDM are present; otherwise the existing WebKit preview fallback still applies.
- **Persistent profile + one-time warm-up.** This was the non-obvious part. A system Chromium registers its preinstalled Widevine CDM through the component-updater, which writes a "hint file" into the *profile* on first launch that only takes effect on a subsequent launch. So the arm64 path launches via a persistent user-data-dir (`~/.cache/vibez/chromium-arm64`) and warms the CDM up once during `EnsureBrowser`. An ephemeral profile (the default `Launch`) would never load Widevine.
- The amd64-only `vibez-helper` hard link and explicit `--widevine-path` are skipped on arm64 (Chromium locates its registered CDM itself).

## Verification

Tested on an Apple Silicon MacBook running Arch Linux ARM (aarch64, glibc 2.43), Chromium 150, with the `widevine` package (4.10.2891.0) installed:

- An EME check — `navigator.requestMediaKeySystemAccess('com.widevine.alpha', …)` — resolves and creates MediaKeys, with a ClearKey control also passing.
- A real Apple Music track plays full-length (well past the 30-second preview boundary) through the CDP backend.
- `make build`, `go vet ./...`, and the full `go test ./...` suite pass. The `cdp` tests are split by arch build tag (`browser_amd64_test.go` / `browser_arm64_test.go`) with new arm64 coverage for discovery, `Available()`, and launch args.

## Requirements for arm64 users

- A system Chromium and a Widevine CDM — for example `pacman -S chromium widevine` on Arch Linux ARM (glibc ≥ 2.36).
- Build from source with `make build`.

## Follow-ups (intentionally not in this PR)

- Publishing prebuilt arm64 release binaries. The Linux build is `CGO_ENABLED=1` (webkit), so it needs an aarch64 cross-toolchain or a native arm64 runner; `.goreleaser.yml` and CI still produce amd64 only, with the reason documented inline. Happy to follow up on arm64 artifacts if that's wanted.
- Flatpak `aarch64`.

Feedback very welcome on the approach — especially the "discover a system Chromium" model versus a fully self-contained download. The latter isn't currently practical on arm64 (there's no Google arm64 Chrome, and the Widevine CDM can't be cleanly bundled), but I'm glad to adjust if you'd prefer a different direction.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -21,6 +21,11 @@ builds:
       - linux
     goarch:
       - amd64
+      # linux/arm64 works at runtime (internal/player/cdp discovers a system
+      # Chromium + Widevine CDM), but is not published here yet: this CGO/webkit
+      # build needs an aarch64 toolchain (CC=aarch64-linux-gnu-gcc + arm64
+      # webkit2gtk) or a native arm64 runner. Until then arm64 users build from
+      # source with `make build`.
 
   - id: vibez-darwin
     binary: vibez
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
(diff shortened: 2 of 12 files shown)


simonepelosi:
That's amazing! Please fix the conflicts and I will proceed with the merge!

PR #95, opened 2026-07-17
Fix radio mode when seeded from a single library song; larger refill batches

ianswope:
## Summary

Radio mode (`R`) worked when seeded from a playlist but failed when seeded from a single library song, and each refill only pulled ~2 tracks. This fixes the library-song seeding and gives refills more runway.

## Root cause

Verified against the live API: an Apple Music per-song station (`ra.<id>`) only accepts a **catalog** song ID. Seeding with a **library** ID (`i.XXXX`) returns HTTP 500 ("Service failure: Stations"). Playlist tracks carry catalog IDs, so playlist-seeded radio worked; a library single song's ID is `i.XXXX`, so `GetStationTracks` sent `ra.i.XXXX` and the station failed.

## Changes

- `GetStationTracks` resolves a library seed to its catalog ID before seeding the station, using the song's `playParams.catalogId` from `GET /me/library/songs/{id}`. When a library song has no catalog equivalent (e.g. an uploaded/unmatched track), it returns a clear error instead of a doomed 500.
- Because `next-tracks` returns only ~2 tracks per call, `GetStationTracks` now pages a few times (deduped, up to a target of 10) so refills give the queue meaningful runway.
- Added provider tests: multi-page batching, library-ID → catalog-ID resolution, and the no-catalog error path.

## Verification

- Live API: `ra.<catalogId>` → 200 with tracks; `ra.i.<libraryId>` → 500; resolving the library song's `playParams.catalogId` then seeding → 200. Repeated `next-tracks` calls return fresh tracks (10 unique across 5 calls), which is what the batching relies on.
- `make build`, `go vet ./...`, the full `go test ./...` suite, and `golangci-lint` are all clean.

## Not included (follow-up)

There's a separate issue where radio can stop advancing after a track — a client-side auto-advance stall in the MusicKit JS queue (`playbackStateDidChange` → "completed" handler), unrelated to seeding. That needs runtime logs to pin down and will be a separate change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- **Radio refills pull a larger batch** — Apple's `next-tracks` endpoint returns only ~2 tracks per call, so `GetStationTracks` now pages several times (deduped, up to a target of 10) to give the queue more runway between refills.
+
 ### Fixed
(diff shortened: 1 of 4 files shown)


simonepelosi:
Please solve conflicts inside the CHANGELOG

Issue #93, opened 2026-07-17
[Bug] Apple Music discovery mode fails with "[vibe] search error: no results"


LazyGamer1111:
Hi, just installed vibez and it's a really cool app, but the discovery feature doesnt work for me. Vibe searching barely works and only finds a few songs which can be wildly different, but discovery mode just simply finds no songs which results in an empty queue, no matter which similarity mode i choose.
### What works:
- playlist loading
- song playing
- authentication
- radio mode
- ...
### What barely works:
- Vibe search only finds a few songs and they are wildly different most of the time
### What doesn't work:
- Discovery mode
### Logs:
```text
19:50:24 [discovery] metric set: 75% similarity
19:50:24 [discovery] started from "If You See Me Use Ubuntu" (similarity 75%, mode=auto)
19:50:24 [discovery] queries (sim=75%): [Grant Knows Weyes Blood Sufjan Stevens]
19:50:24 [vibe] search error: no results
19:50:24 [discovery] queries (sim=75%): [Grant Knows The National Sufjan Stevens]
19:50:25 [vibe] search error: no results
19:50:52 [discovery] stopped
19:50:54 [discovery] metric set: 10% similarity
19:50:54 [discovery] started from "If You See Me Use Ubuntu" (similarity 10%, mode=auto)
19:50:54 [discovery] queries (sim=10%): [The Tallest Man on Earth Tyler Childers Josh Ritter]
19:50:55 [vibe] search error: no results
19:50:55 [discovery] queries (sim=10%): [Vince Staples Dustin O'Halloran Kendrick Lamar]
19:50:56 [vibe] search error: no results

```
### Environment:
- Provider: Apple Music

ianswope:
Thanks for the logs, they made this easy to find. I can reproduce it, and it isn't specific to your account or library.

`Search` runs three requests in parallel (apple.go:363): library, catalog songs, and catalog albums/playlists. Only the catalog-songs leg uses `amp-api.music.apple.com`, and that host is currently returning 401 to everything, with an empty body. The same token pair gets 200 from `api.music.apple.com` for the identical catalog search, and 200 from `me/library/search`, so the tokens themselves are fine.

The failure is invisible because `Search` only returns an error if all three legs fail (apple.go:461). Library and albums/playlists succeed, so you get library-only results with a nil error. Discovery then throws its own error away too (model.go:2488), so nothing ever reaches the log.

[...]

PR #98, opened 2026-07-27
fix(search): surface partial provider failures instead of empty results

ianswope:
Refs #93. This is the diagnostics half of that issue, deliberately separate from the fix for the underlying breakage, since that one needs a decision from you about the `extendedAssetUrls` filter.

## The problem

`Search` fans out to three parallel requests — library, catalog songs, catalog albums/playlists — and only returns an error when all three fail (apple.go:461). One dead backend therefore produces a thinner result set with a nil error, and no caller can tell the difference between that and a query nothing matched.

The discovery refill made it worse by dropping its error on the floor entirely (`if err != nil || res == nil { c <- out{} }`), which is why the report in #93 contains nothing more useful than `[vibe] search error: no results` despite the catalog-song backend returning 401 on every request.

## What this changes

- `provider.SearchResult` gains a `Warnings []string` field for non-fatal backend failures.
- `apple.Search` records which leg failed instead of quietly returning less.
- The vibe search and discovery refill collect failures and warnings rather than discarding them, deduplicated because parallel queries against one provider report the same backend failure once per query.
- Warnings are logged as `[search] partial results: …` on both the plain search and the vibe/discovery paths, and named in the `no results` error so the log line carries a cause.

Results themselves are unchanged. A failing leg still contributes nothing, the other legs are still returned, and the success path behaves exactly as before. The vibe panel still shows its usual `no results` state, since it renders a fixed string rather than the error text.

With this applied, the situation in #93 logs as:

```
[search] partial results: catalog song search: apple music api 401 Unauthorized. Re-run `vibez login`; if you're using a local build, also refresh apple_developer_token.
[vibe] search error: no results (catalog song search: apple music api 401 Unauthorized. …)
```

## Tests

- `TestSearch_PartialFailureRecordedAsWarning` — 401 on the catalog-song leg only; asserts a nil error, the library track still returned, and one warning naming the failing leg.
- `TestSearch_NoWarningsWhenAllLegsSucceed` — an empty result is not a failure.
- `TestDedupeStrings`, `TestNoResultsError` — helper behaviour, including that a repeated reason collapses to one.

`go build ./...`, `go vet ./...`, `go test ./...` and `golangci-lint run` all pass.

## The first commit

7fed29f is a one-line tooling fix and unrelated to the rest — happy to pull it into its own PR if you'd rather.

The `musickit-js-test` hook declares `language: node`, which asks pre-commit to build a Node environment by installing the local package. With no `package.json` and no `additional_dependencies`, pre-commit 4.x aborts during environment installation:

```
AssertionError: `language: node` must have package.json or additional_dependencies
```

That fires at install time for every hook in the config, so it blocks any commit regardless of whether the staged files match the hook's own `files` pattern. It's why I couldn't commit this branch until it was fixed.

To be clear about the scope: CI is unaffected today — the `pre-commit` job is green on the other open PRs — so this is a local-environment failure rather than a broken pipeline. It'll reach anyone whose `pre-commit` is current, though, which is why it seemed worth a line rather than a local workaround.

The script has no npm dependencies and only needs a node binary, so `language: system` is the right declaration and matches the four Go hooks above it.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -46,7 +46,7 @@ repos:
     hooks:
       - id: musickit-js-test
         name: musickit js test
-        language: node
+        language: system
         entry: node internal/player/web/musickit_test.mjs
         files: '^internal/player/(web|webkit/web)/musickit'
         pass_filenames: false
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **Search failures were reported as empty results** — `Search` queries the library, catalog songs, and catalog albums/playlists in parallel and only returned an error when all three failed, so one dead backend silently produced a thinner result set instead of a failure. Discovery refills then discarded the error entirely, leaving `[vibe] search error: no results` as the only clue. Partial failures are now carried on `SearchResult.Warnings`, logged as `[search] partial results: …`, and named in the "no results" message, so an unreachable backend is distinguishable from a query that genuinely matched nothing. Refs #93.
+
 ## [0.5.0] — 2026-07-10
 
 ### Added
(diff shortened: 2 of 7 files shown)


simonepelosi:
Hey! I'm back, sorry I was on vacation! That seems great to me! I would approve and merge!

simonepelosi (approved):

Issue #88, opened 2026-07-10
The app closes out of the blue


unlikeghost:
I'm not sure what details to provide, but the app has suddenly started closing on its own. No logs are appearing in the debug output, and this behavior began immediately after updating to the latest version.

- OS: Omarchy 3.8.2 (Arch Linux derivative)
- Kernel: Linux 6.18.9-arch1-1-g14
- Compositor: Hyprland 0.55.2 (Wayland)
- Terminal: Ghostty 1.3.1-arch2
- CPU: AMD Ryzen 7 5800HS
- GPU: NVIDIA GeForce RTX 3060 Mobile / AMD Radeon Vega

simonepelosi:
Hey! So the app is crashing without any issue ? Nothing on the terminal ?

simonepelosi:
So are you using the 0.5.0 ?

unlikeghost:
Yes to both. I usually use a desktop shortcut, but I'll run it from the terminal from now on to check for log outputs

ianswope:
Thanks for the details, and sorry this sat for a while. I can reproduce a mechanism here that matches your description closely, and there are two candidates worth separating before anyone changes code.

**Why the timing fits.** v0.5.0 bumped the Playwright dependency: v0.4.x expected driver `1.60.0`, v0.5.0 expects `1.61.1`. So the first launch after updating is the one launch where `EnsureBrowser` finds a stale driver and runs `playwright.Install()` (`internal/player/cdp/browser.go:92`). Every v0.4.x user hits that exactly once — and since auto-update is on by default, the update itself may well have happened without you asking for it.

**What that does to the terminal.** On that path the node driver process inherits vibez's stdout/stderr — the live TTY that Bubble Tea has already switched to raw mode. node snapshots the terminal attributes when it initialises a TTY stream and re-applies them from its exit handler, and vibez waits for the driver to exit during shutdown (`cmd/cdp_common.go`), so node's restore lands *after* Bubble Tea has restored cooked mode. Measured on Arch + Hyprland here: once the driver exits, the tty is left `-echo -icanon -isig` — no echo, no line editing, and ctrl+c dead. From a `Terminal=true` desktop entry that is hard to tell apart from "the app closed on its own", and it also explains the empty debug output, since the TUI's own log never becomes reachable.

[...]

PR #99, opened 2026-08-05
fix(cdp): keep the playwright driver off the user's terminal

ianswope:
`playwright.Run` starts the node driver through playwright-go's `newPipeTransport`, which pipes the child's stdin and stdout but passes `RunOptions.Stderr` through untouched — and that defaults to `os.Stderr`, a live TTY, while Bubble Tea holds the terminal in raw mode. Node snapshots the terminal's termios when it initialises a TTY stream and re-applies that snapshot from its exit handler. Shutdown stops the driver (`cmd/cdp_common.go`), so the restore lands *after* Bubble Tea has returned the terminal to cooked mode, silently undoing it.

The shell is left with `-echo -icanon -isig`: no echo, no line editing, and ctrl+c dead until the window is closed.

Measured on Arch + Hyprland, reproducing the ordering vibez uses (raw mode → driver runs → cooked mode restored → driver exits). Only stderr varies; stdin and stdout are pipes in both arms, as `newPipeTransport` always makes them:

| driver stderr | after the TUI restores | after the driver exits |
| --- | --- | --- |
| inherited TTY (today) | `echo icanon isig` | `-echo -icanon -isig` |
| piped (this PR) | `echo icanon isig` | `echo icanon isig` |

**Who reaches it:** every session that starts full-track playback, on Linux and macOS both. An earlier revision of this PR described it as an upgrade-only path that ran through `playwright.Install()`; that was wrong, and the real scope is wider.

**The fix:** route the driver's stderr to `io.Discard` at both `playwright.Run` call sites — `runPlaywright` in `browser.go` and `browser_darwin.go`. Returning a plain `io.Writer` rather than an `*os.File` is load-bearing: `os/exec` only substitutes a pipe when the writer isn't a file, and that is what keeps the driver away from the terminal. The test guards that property rather than the specific writer, so routing driver diagnostics into a log later won't be a false failure.

`RunOptions.Logger` goes in alongside it. `transformRunOptions` calls `log.SetOutput(RunOptions.Stderr)` whenever `Stderr` is set and `Logger` is not, so setting `Stderr` on its own would send the whole process's standard log output to `io.Discard` as a side effect. Credit to @elijah-rou, who identified this in #103.

**Deliberately not touched:** the `playwright.Install` path. Every call site passes `SkipInstallBrowsers`, so `Install` returns after `DownloadDriver` and never reaches `installBrowsers` — the only place that assigns `RunOptions.Stdout` to a child. The one process it does start, the `--version` check, runs through `exec.Cmd.Output`, which captures both streams into buffers.

This may be part of #88 (app closing with no output, starting right after the update to v0.5.0). I'm still waiting on the reporter's exit code and terminal state to confirm that particular case, so this is `Refs #88` rather than a close — the terminal corruption is worth fixing either way.

Verification: `make test`, `make lint` (0 issues), `go test -race -count=1 ./internal/player/cdp`, `go vet`, plus cross-builds of `internal/player/cdp` for `darwin/{arm64,amd64}` and `linux/{arm64,amd64}`.


ianswope:
Rebased onto current `main` and force-pushed — head is now `7719133`. @elijah-rou this is your cue to rebase #103, and the old head this was stacked on is gone, so the extra commit should drop out of your diff.

Worth flagging what the rebase did *not* do on its own, since it's the thing I raised on #103 and it turned out to bite here first.

#94 split `EnsureBrowser` in `browser.go` into `ensureBrowserARM64` and `ensureBrowserAMD64`, so there are now two `playwright.Install` call sites where there was one. The rebase merged **cleanly** — no conflict, `mergeable` stayed green — and applied the hunk to the amd64 side only, leaving `ensureBrowserARM64` still handing node a live TTY. Nothing in the PR view would have shown it. I test-merged into `main` to check rather than trusting the mergeable flag, which is the only reason I caught it.

So the fix now covers every site that starts the driver:

- `browser.go` — `ensureBrowserARM64`, `ensureBrowserAMD64`, `runPlaywright`
- `browser_darwin.go` — `EnsureBrowser`, `runPlaywright`

`isDriverUpToDate` also sets them, which is harmless but unnecessary — it only reads `driver.Version` and never starts a process. #103 removes that, correctly.

The arm64 path is the one I actually run, so it would have shipped a terminal fix that didn't work on the machine I use to test terminal fixes.

Verified: `make test`, `make lint` (0 issues), `go vet ./internal/player/cdp/`, and clean cross-compiles for `darwin/arm64` and `linux/arm64`.

ianswope:
Rebased onto `d62f9f6` — head is now `504dcae`, and the PR is smaller than it was. With a correction to my last comment.

#100 moved both `playwright.Install` call sites and `isDriverUpToDate` into `driver.go`, which conflicted with three of the five sites this PR patched. Resolving that meant reading playwright-go's driver properly, and it turns out those three hunks were doing nothing. Including the arm64 one I made a point of last time — the "would have shipped a terminal fix that didn't work" framing was chasing a hunk that was inert to begin with. Sorry for the noise.

**What actually causes it.** With `SkipInstallBrowsers: true`, which every call site here passes, `playwright.Install` never hands node a terminal:

- `Install` → `DownloadDriver`, then returns before `installBrowsers` — the only function that assigns `cmd.Stdout = options.Stdout`
- `DownloadDriver` → `isUpToDateDriver` → `d.Command("--version").Output()`. `Command` sets neither stream, and `Output` captures both into buffers
- everything else on that path is HTTP and file I/O

The leak is `playwright.Run` → `driver.run` → `newPipeTransport(d, d.options.Stderr)`, which pipes the child's stdin and stdout but sets `cmd.Stderr = os.Stderr`. That is the long-lived `run-driver` process, stopped during shutdown via `cmd.Wait()` — exactly the ordering in the symptom. Both `runPlaywright` sites, and only those, need the fix. #100 left them untouched, so this merges cleanly now.

I re-measured rather than trusting the read, varying only stderr and leaving stdin/stdout as pipes in both arms:

| driver stderr | after the TUI restores | after the driver exits |
| --- | --- | --- |
| inherited TTY (today) | `echo icanon isig` | `-echo -icanon -isig` |
| piped (this PR) | `echo icanon isig` | `echo icanon isig` |

So stderr alone is sufficient to cause it, which also means the bug is **wider** than I described, not narrower: it is not once-per-user-on-upgrade, it is every session that starts full-track playback, on Linux and macOS both. I've corrected the PR body, which had it as an upgrade-only path.

**One addition.** Setting `RunOptions.Stderr` without `Logger` makes `transformRunOptions` call `log.SetOutput(...)`, so this fix as written would have pointed the whole process's standard logger at `io.Discard` as a side effect. @elijah-rou spotted that in #103, and I've folded the one-line half of it in here — this PR is what introduces the side effect, and it shouldn't sit in `main` waiting on a stacked draft. The substantive half of #103 is untouched and still yours: the ring buffer that keeps the newest node output and attaches it to startup errors is the part worth reviewing, and it's the thing that would have made #88 diagnosable in the first place. You'll need a rebase either way now that #100 has landed.

There's a regression test for the logger branch that exercises `playwright.NewDriver` for real rather than asserting on my reading of it — it fails if the `Logger` field goes away or playwright-go moves the branch. Verified by deleting the field and watching it fail.

Verification: `make test`, `make lint` (0 issues), `go test -race -count=1 ./internal/player/cdp`, `go vet`, plus `internal/player/cdp` builds for `darwin/{arm64,amd64}` and `linux/{arm64,amd64}` and test-binary compiles for `darwin/arm64` and `linux/arm64`.

simonepelosi:
This LGTM and it looks like a great improvement for terminal stability!

simonepelosi (approved):

Issue #89, opened 2026-07-13
version 0.5.0 - Playwright driver error


mattsaund:
Updated to version 0.5.0

"browser setup: playwright driver: could not install driver: could not install driver: driver exists but version not 1.61.1 in : [filepath]"

when I go to play songs it then crashes "program was killed: program experienced a panic"

specs:
OS: Linux mint cinnamon 22.3
CPU: 12th Gen Intel Core i5-12400
GPU: NVIDIA GeForce RTX 4070
RAM: 32GB DDR4 3000

Notes:
Previous version worked fine before updating.

simonepelosi:
Mhhh that seems a cache issue, did you tried to remove the Vibez cache folder ?

mattsaund:
> Mhhh that seems a cache issue, did you tried to remove the Vibez cache folder ?

ill give that a try soon and report back

mattsaund:
Clearing cache fixed this issue!

simonepelosi:
Thank you so much ❤️

PR #100, opened 2026-08-06
Fix Playwright driver cache upgrades

elijah-rou:
## Summary
- remove Vibez's stale Playwright driver cache before installing a newly bundled driver version
- validate the cached CLI's actual version so incomplete or inconsistent caches are also repaired
- preserve the separate cached Chrome installation
- share the upgrade handling across Linux and macOS

Refs #89.

## Tests
- `uvx pre-commit run --all-files`
- `PKG_CONFIG_PATH="$PWD/pkg-config" go test -count=1 ./...`
- Darwin amd64 and arm64 CDP test cross-compilation
- Windows amd64 CDP test cross-compilation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+- **Playwright driver upgrades failed against an older cache** — Remove Vibez's cached Playwright driver before installing a different bundled version. This prevents upgrades from failing with `driver exists but version not ...`; the separately cached Chrome installation is preserved. Refs #89.
 - **Search failures were reported as empty results** — `Search` queries the library, catalog songs, and catalog albums/playlists in parallel and only returned an error when all three failed, so one dead backend silently produced a thinner result set instead of a failure. Discovery refills then discarded the error entirely, leaving `[vibe] search error: no results` as the only clue. Partial failures are now carried on `SearchResult.Warnings`, logged as `[search] partial results: …`, and named in the "no results" message, so an unreachable backend is distinguishable from a query that genuinely matched nothing. Refs #93.
 
 ## [0.5.0] — 2026-07-10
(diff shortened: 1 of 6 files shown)


ianswope:
First-pass review note rather than a verdict — flagging a collision before anyone
burns time on it.

**This and #99 conflict.** Both touch `internal/player/cdp/browser.go` and
`browser_darwin.go`, and `git merge-tree` reports content conflicts in both. It's the
same lines by coincidence of purpose: #99 adds `Stdout`/`Stderr` to the
`playwright.RunOptions` literals, and this PR moves those literals into
`installPlaywrightDriver()` in the new `driver.go`.

**They're complementary, not competing** — the ordering only decides who absorbs the
rebase:

- If **this lands first**, #99 gets *smaller*: instead of patching four separate
  `RunOptions` literals, it sets the writers once in `driver.go`. Strictly better end
  state.
- If **#99 lands first**, this PR's refactor has to absorb the conflict, which is the
  more annoying direction for no benefit.

So my suggestion is **#100 → #99 → #103**, and I'll do the #99 rebase myself once
this merges — @elijah-rou, nothing for you to pick up here.

Worth noting #103 also touches `isDriverUpToDate`, which this PR relocates, so it
wants to be last regardless of what happens to the first two.

@simonepelosi this is a merge-ordering call rather than a code one — all three are
sound independently, and none of them blocks the others on correctness.

elijah-rou:
I'm personally ez, lets do the 100 -> 99 -> 103 order @ianswope suggested

simonepelosi:
Hey! Sorry again for my late reviews! Can you solve the conflicts inside the browser_test file ? @elijah-rou

elijah-rou:
> Hey! Sorry again for my late reviews! Can you solve the conflicts inside the browser_test file ? @elijah-rou

but of course, gonna do it right now

elijah-rou:
Resolved in two commits:

- `22ca4ce` merges current `main` and resolves the `browser_test.go` split by keeping the new architecture-specific test files and carrying the executable-driver fixture into the amd64 test.
- `f9d054d` routes the new Linux/arm64 install path through the shared upgrade-safe installer, covering the extra Playwright install site introduced by #94 that Ian called out.

simonepelosi:
It looks great!

PR #101, opened 2026-08-06
fix(mpris): harden track identity updates

elijah-rou:
## Summary

- generate bounded, collision-resistant MPRIS track paths from the primary ID, catalog ID, or length-prefixed metadata
- contain godbus property panics and disable MPRIS cleanly if the session bus is lost
- serialize debounced publication with shutdown and reject pending or post-close updates
- validate `SetPosition` against the current track and position bounds

## Why

An empty MusicKit track ID produces `/org/vibez/track/`. godbus rejects that path while metadata is flushed and `SetMust` panics in the timer goroutine, terminating vibez. Lossy path sanitization can also conflate distinct tracks, while bus loss or a shutdown race can trigger the same panic-on-error API.

## Tests

- `go test -count=1 ./...`
- `go vet ./...`
- targeted MPRIS regression tests under `go test -race -count=10`

### DIFF
--- a/internal/player/mpris/mpris_test.go
+++ b/internal/player/mpris/mpris_test.go
@@ -775,13 +775,132 @@ func TestServer_Update_TrackWithArtwork(t *testing.T) {
 	srv.Update(player.State{Playing: true, Track: track})
 }
 
+func TestServer_Update_TrackWithoutID(t *testing.T) {
+	ctrl := &mockController{}
+	srv, err := NewServer(ctrl)
+	if err != nil {
+		t.Skipf("NewServer failed: %v", err)
+	}
+	defer srv.Close() //nolint:errcheck
+
+	track := &provider.Track{Title: "Partial Track"}
+	srv.Update(player.State{Playing: true, Track: track})
+	time.Sleep(300 * time.Millisecond)
+
+	metadata, ok := srv.props.GetMust(mprisPlayerIface, "Metadata").(map[string]dbus.Variant)
+	if !ok {
+		t.Fatal("Metadata property has unexpected type")
+	}
+	trackID, ok := metadata["mpris:trackid"].Value().(dbus.ObjectPath)
+	if !ok {
+		t.Fatal("mpris:trackid has unexpected type")
+	}
+	want := trackObjectPath(track)
+	if trackID != want {
+		t.Fatalf("mpris:trackid = %q, want %q", trackID, want)
+	}
+}
+
 func TestServer_Close(t *testing.T) {
 	ctrl := &mockController{}
 	srv, err := NewServer(ctrl)
 	if err != nil {
 		t.Skipf("NewServer failed: %v", err)
 	}
-	if err := srv.Close(); err != nil {
+	srv.Update(player.State{Playing: true, Track: &provider.Track{ID: "pending"}})
+
+	// Holding flushMu models a callback already publishing. Close must mark the
+	// server closed immediately, then wait before closing the D-Bus connection.
+	srv.flushMu.Lock()
+	closeDone := make(chan error, 1)
+	go func() { closeDone <- srv.Close() }()
+	deadline := time.Now().Add(time.Second)
+	for {
+		srv.mu.Lock()
+		closed := srv.closed
+		srv.mu.Unlock()
+		if closed {
+			break
+		}
+		if time.Now().After(deadline) {
+			srv.flushMu.Unlock()
+			t.Fatal("Close did not mark the server closed")
+		}
+		time.Sleep(time.Millisecond)
+	}
+	select {
+	case err := <-closeDone:
+		srv.flushMu.Unlock()
+		t.Fatalf("Close returned before the in-flight flush completed: %v", err)
+	default:
+	}
+	srv.flushMu.Unlock()
+	if err := <-closeDone; err != nil {
 		t.Errorf("Close() error: %v", err)
 	}
+
+	// Pending and post-close updates must not flush against the closed D-Bus connection.
+	srv.Update(player.State{Playing: true, Track: &provider.Track{ID: "closed"}})
+	time.Sleep(200 * time.Millisecond)
+}
+
+func TestServer_Update_DistinguishesCollidingTrackIDs(t *testing.T) {
+	ctrl := &mockController{}
+	srv, err := NewServer(ctrl)
+	if err != nil {
+		t.Skipf("NewServer failed: %v", err)
+	}
+	defer srv.Close() //nolint:errcheck
+
+	first := &provider.Track{ID: "i.track/1", Title: "First"}
+	srv.Update(player.State{Playing: true, Track: first})
+	srv.flush()
+
+	second := &provider.Track{ID: "i_track_1", Title: "Second"}
+	srv.Update(player.State{Playing: true, Track: second})
+	srv.flush()
+
+	metadata, ok := srv.props.GetMust(mprisPlayerIface, "Metadata").(map[string]dbus.Variant)
+	if !ok {
+		t.Fatal("Metadata property has unexpected type")
+	}
+	title, ok := metadata["xesam:title"].Value().(string)
+	if !ok {
+		t.Fatal("xesam:title has unexpected type")
+	}
+	if title != second.Title {
+		t.Fatalf("xesam:title = %q, want %q", title, second.Title)
+	}
+}
+
+func TestServer_UpdateAfterBusLossDoesNotPanic(t *testing.T) {
+	ctrl := &mockController{}
+	srv, err := NewServer(ctrl)
+	if err != nil {
+		t.Skipf("NewServer failed: %v", err)
+	}
+
+	srv.Update(player.State{Playing: true, Track: &provider.Track{ID: "track"}})
+	if err := srv.conn.Close(); err != nil {
+		t.Fatalf("close D-Bus connection: %v", err)
+	}
+
+	// flush runs synchronously so any panic fails this test directly.
+	srv.flush()
+
+	srv.mu.Lock()
+	unavailable := srv.unavailable
+	srv.mu.Unlock()
+	if !unavailable {
+		t.Fatal("server remained available after losing its D-Bus connection")
+	}
+
+	srv.Update(player.State{Playing: true, Track: &provider.Track{ID: "ignored"}})
+	srv.mu.Lock()
+	pending := srv.pending
+	srv.mu.Unlock()
+	if pending != nil {
+		t.Fatal("server accepted an update after losing its D-Bus connection")
+	}
+	_ = srv.Close()
 }
--- a/internal/player/mpris/mpris_test.go
+++ b/internal/player/mpris/mpris_test.go
@@ -785,7 +785,7 @@ func TestServer_Update_TrackWithoutID(t *testing.T) {
 
 	track := &provider.Track{Title: "Partial Track"}
 	srv.Update(player.State{Playing: true, Track: track})
-	time.Sleep(300 * time.Millisecond)
+	srv.flush()
 
 	metadata, ok := srv.props.GetMust(mprisPlayerIface, "Metadata").(map[string]dbus.Variant)
 	if !ok {
--- a/internal/player/mpris/server.go
+++ b/internal/player/mpris/server.go
@@ -91,7 +91,10 @@ func trackObjectPath(track *provider.Track) dbus.ObjectPath {
 			_, _ = hash.Write(fieldLength[:])
 			_, _ = hash.Write([]byte(field))
 		}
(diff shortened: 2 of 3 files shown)


ianswope:
First-pass review. **The bug is real and I reproduced it** — I'd like to see this
land. One thing blocks it, and one thing isn't yours.

## The premise checks out

I verified both halves rather than taking the description's word for it:

```
path="/org/vibez/track/" IsValid=false
PANIC REPRODUCED from SetMust: dbus: wire format error: invalid object path
```

An empty `Track.ID` really does produce an invalid object path, and publishing it
through `props.SetMust` really does panic. Since `flush` runs from a
`time.AfterFunc` timer goroutine, that panic is unrecoverable and takes the whole
process down — a media-key press at the wrong moment kills the player. Worth fixing.

The length-prefixed hash for the no-ID case is the right call too: without the
length prefix, `{Title: "ab", Artist: "c"}` and `{Title: "a", Artist: "bc"}` would
collide.

## Blocking: this currently fails CI's lint gate

`golangci-lint v2.11.4` (the version pinned in `.pre-commit-config.yaml`, which the
`pre-commit` CI job runs) reports **3 issues on this branch and 0 on `main`**:

```
internal/player/mpris/server_test.go:164:19: Error return value of `obj.setPosition` is not checked (errcheck)
internal/player/mpris/server_test.go:178:17: Error return value of `obj.setPosition` is not checked (errcheck)
internal/player/mpris/server.go:94:52: G115: integer overflow conversion int64 -> uint64 (gosec)
```

The two test ones are just `_ = obj.setPosition(...)`. For G115, `track.Duration`
is an `int64` going into `PutUint64` — a negative duration would wrap to an
enormous value. It only feeds a hash so nothing breaks, but it needs either an
explicit guard or a `//nolint:gosec` with that reasoning.

**Related and worth knowing:** GitHub reports *no checks at all* on this PR, and
the same on #100 — while #103 ran the full suite and passed. So the lint failure
above wouldn't have been caught by CI as things stand. @simonepelosi that's
probably a workflow-approval gate on fork PRs rather than anything either of us
did, but it means these two PRs currently carry no CI signal.

## Not your bug: the `-race -count=10` claim

The description lists `go test -race -count=10` under Tests, but that command fails
on this branch — and it fails identically on `main`, so **you didn't introduce it**:

```
--- FAIL: TestMPRISPlayer_ReadState_WithMetadata
    testing.go:1712: race detected during execution of test
```

It's entirely in the test harness: `fakeMPRIS.Get` (`mpris_test.go:374`) is served
on godbus's inbound worker goroutine while the test body writes those same fields
at `mpris_test.go:646`, with the object already exported on the bus. Product code
isn't involved. Opened as #107 — flagging it only because it's
cited as passing evidence here.

## Minor: `trackObjectPath` panics

```go
if !path.IsValid() {
    panic(fmt.Sprintf("mpris: generated invalid track path %q", path))
}
```

This is unreachable today — the namespace is one of three literals and the rest is
hex — so it reads as an assertion. But it sits slightly against the PR's own
thesis: `trackObjectPath` is called from `Update`, on the audio-engine subscriber
goroutine, outside the `trySetProperty` recover. If it ever did fire it would kill
vibez in exactly the way this PR exists to prevent. Returning `noTrackPath` would
keep the same intent without the sharp edge.

## Things I checked that are fine

- **Double-close.** `markUnavailable` closes the conn and `Close()` can close it
  again, but godbus guards with `closeOnce`, so that's safe.
- **Lock ordering.** `Close` releases `mu` before taking `flushMu`, and `flush`
  takes `flushMu` then `mu` — no inversion, no deadlock.
- **`SetPosition` bounds.** `maxPositionUs` is derived so the later
  `time.Duration(posUs) * time.Microsecond` can't overflow. Correct.
- **Dropping a seek that races a track change** is the right trade — honouring it
  would apply the old track's position to the new one, which is worse than a no-op.

Fix the three lint findings and I'm happy to see this go in. Nice catch on the
underlying crash.

ianswope:
Re-checked this against current `main` (`d62f9f6`) now that #100 and #104 have landed. Status is unchanged and good: **it still merges cleanly** — five merges have gone in since this branch's base and none of them touch `internal/player/mpris`, so there's no rebase pressure on you.

The three lint findings from my last pass all still reproduce under the pinned `golangci-lint v2.11.4`. Rather than just re-flagging them, here's the patch — verified at `0 issues`, with `go test ./internal/player/mpris` and `go vet` still clean:

```diff
--- a/internal/player/mpris/server.go
+++ b/internal/player/mpris/server.go
@@ -91,7 +91,10 @@ func trackObjectPath(track *provider.Track) dbus.ObjectPath {
 			_, _ = hash.Write(fieldLength[:])
 			_, _ = hash.Write([]byte(field))
 		}
-		binary.BigEndian.PutUint64(fieldLength[:], uint64(track.Duration))
+		// int64 -> uint64 reinterprets the bits rather than losing them, so the
+		// hash stays injective over Duration. Clamping a negative value would
+		// instead fold distinct tracks onto the same digest.
+		binary.BigEndian.PutUint64(fieldLength[:], uint64(track.Duration)) //nolint:gosec // bijective conversion; hash input only
 		_, _ = hash.Write(fieldLength[:])
 		copy(sum[:], hash.Sum(nil))
 	} else {
--- a/internal/player/mpris/server_test.go
+++ b/internal/player/mpris/server_test.go
@@ -161,7 +161,7 @@ func TestPlayerObj_SetPositionValidatesTrackAndPosition(t *testing.T) {
 			obj := &playerObj{ctrl: ctrl, srv: srv}
-			obj.setPosition(test.trackPath, test.position)
+			_ = obj.setPosition(test.trackPath, test.position)
@@ -175,7 +175,7 @@ func TestPlayerObj_SetPositionValidatesTrackAndPosition(t *testing.T) {
 	obj := &playerObj{ctrl: ctrl, srv: srv}
-	obj.setPosition(currentPath, int64((5 * time.Second).Microseconds()))
+	_ = obj.setPosition(currentPath, int64((5 * time.Second).Microseconds()))
```

On G115 I went with a `//nolint` rather than the guard I suggested last time, because I had it backwards: clamping a negative `Duration` to `0` would be the *lossy* option. `uint64(int64)` reinterprets the bits, so it's a bijection and the digest stays injective over `Duration` — a clamp would fold every negative value onto the same hash input, which works against the collision-resistance this function exists for. Wrapping is the right behaviour here; gosec just can't tell hash input from arithmetic.

`_ =` is right for the two test sites rather than asserting the return, since `setPosition` returns `nil` on every path — a stale-path SetPosition is a deliberate silent no-op, which matches the spec.

**Still no CI signal on this PR.** `gh pr checks 101` reports no checks at all, while #103 from the same fork ran the full suite and passed. @simonepelosi I think that's a per-PR "Approve and run workflows" click on your side rather than anything wrong with the branch — worth doing before merge, since the lint gate above is exactly what CI would have caught.

The `-race -count=10` failure is still #107 and still not yours: I re-ran it on current `main` and it reproduces identically there. Every frame in the race report is in `mpris_test.go` (the `fakeMPRIS.Get` harness), with no `server.go` on either stack.

With the lint patch applied I'm happy with this. The underlying crash is real, the length-prefixed hash is the right shape, and the panic-containment around `SetMust` is worth having on its own.

ianswope (commented):
Read through this against current `main` and ran it locally. This is careful work and the core of it is clearly right — thanks for picking it up.

I reproduced the crash on `main` to be sure of what we're fixing: `go test -race -count=10 ./internal/player/mpris/...` panics with `dbus: connection closed by user` from `SetMust` at `server.go:339`, inside the debounce timer goroutine (`created by time.goFunc`). Unrecovered and off the main goroutine, so it takes the process with it. On your branch that's gone. `trySetProperty` + `markUnavailable` is the right shape for this — godbus's panic-on-error API doesn't belong anywhere near an optional feature.

The identity work looks right to me too. Length-prefixing the hashed fields is the detail I'd have missed — it's what stops `("ab","c")` and `("a","bc")` colliding — and namespacing `id`/`catalog`/`unknown` keeps the three sources from colliding with each other. The old `sanitizePathElement` conflating `i.track/1` with `i_track_1` is a real bug and your test pins it directly. `setPosition` ignoring a stale `TrackId` also matches what the MPRIS spec asks for.

Two things before I'd call this done:

**1. `-race` fails on `TestServer_Update_DistinguishesCollidingTrackIDs`.** The PR notes `-race -count=10` passing; I get a data race on any machine with a session bus. It's between `trySetProperty("Metadata", meta)` and godbus's outgoing-signal encoder: `prop.set` does `Store([]any{v}, prop.Value)`, which mutates the stored map in place, and `emitChange` passes that same map into `Emit` — so the worker goroutine is encoding the map a later flush is rewriting.

To be clear, this is **not yours** — `main` races the same way in `TestMPRISPlayer_ReadState_*`, so it predates this PR. But your new test is the first thing to trip it reliably, so it'll look like a regression to whoever runs `-race` next. Either hand godbus a fresh map per set so it has nothing shared to mutate, or leave it and say so explicitly in the PR body so it's a known pre-existing issue rather than a surprise. I'm happy to send the fix as a separate PR if you'd rather keep this one focused — your call, it's your patch.

**2. `trackObjectPath` panics on an invalid path.** Given the point of this PR is getting panics out of this file, that one reads oddly. It's unreachable by construction — `namespace` is always a literal and `%x` of a sha256 is always 64 hex chars — so it's an assertion rather than a live risk. I'd return `noTrackPath` instead, consistent with the rest of the change.

One observation, not a request: CI reports 47.3% coverage on this package, where I get 69.9% locally. That suggests some of these paths may not be exercised on the runner, so CI passing is probably weaker evidence than it looks for the bus-loss and shutdown paths. Might be worth confirming.

Nothing blocking from me beyond #1 being either fixed or documented.

ianswope:
Re-checked at `245ddd0`. The three lint findings are gone — pinned `golangci-lint v2.11.4` reports 0 issues, and `go test -count=1 ./...` is green. The `flush()` swap for the 300ms sleep is a nice tightening beyond what I flagged.

The two items from my last review are both still open, and I suspect that push crossed with it rather than answering it.

**The race still reproduces at this head.** `go test -race -count=10 ./internal/player/mpris/...` fails on `TestServer_Update_DistinguishesCollidingTrackIDs` — godbus's encoder calling `reflect.Value.MapKeys` on the metadata map while a later flush writes it.

Worth adding to what I said before: **CI never runs `-race`.** All three `go test` invocations in `ci.yml` (lines 58, 82, 85) are plain. So the green checks aren't weak evidence here, they're no evidence — same direction as the coverage gap I mentioned.

I'd rather not leave you blocked on my question, so let me pick the default: since it's pre-existing on `main`, I'll send the fix as its own PR and this one doesn't need to grow. If you'd rather fold it in, say so and I'll stay out of the file. Either way the **PR body still lists `-race -count=10` under Tests**, and that's the bit worth correcting before Simone reads it — right now it claims a thing that doesn't hold.

`trackObjectPath` still panics at `server.go:105`. Unreachable by construction, so entirely your call, but in a PR whose whole point is getting panics out of this file it'll draw a reviewer's eye.

ianswope:
Correcting myself on one thing above, before you spend time on it: **the fix I suggested wouldn't have worked.**

I said to hand godbus a fresh map per set so it has nothing shared to mutate. It can't help. `prop.Export` allocates the property's storage once at registration (`prop/prop.go:220-222`), and `dbus.Store` merges into that allocation rather than replacing it — `storeMapIntoMap` only calls `MakeMap` when the destination is nil, and otherwise does in-place `SetMapIndex` (`dbus.go:180-195`). `emitChange` then hands `Emit` the *pointer* (`prop/prop.go:300`), and the dereference happens at encode time in godbus's output goroutine (`encoder.go:151`). So the signal always references godbus's single map no matter what we pass in.

The real fix is to stop letting the stored value be the emitted value — register the runtime-updated properties `EmitFalse` and emit `PropertiesChanged` ourselves from a fresh map. That's a design change rather than a patch, so I'll open an issue for it instead of quietly folding it into a side PR. Your PR shouldn't wait on it either way — the race predates you.

Separately, chasing this turned up a pre-existing bug worth its own fix: because `Store` merges and never clears, metadata keys leak across tracks. A track with no artwork keeps the previous track's `mpris:artUrl`, and when `Track` is nil we only write `mpris:trackid`, so stopping leaves the old title/artist/album exported. I'll send that one separately.

simonepelosi:
That's great! I like how this PR hardens  MPRIS D-Bus integration against process panics and desync bugs. It stops invalid or empty track IDs from crashing vibez, safely recovers from D-Bus session bus loss, and prevents stale MPRIS SetPosition calls from triggering out-of-bounds seeks. I really want to merge it and have a new release asap.

simonepelosi (approved):
Great job there!

simonepelosi:
@ianswope I guess we can work on your findings after we merged this, as you said! Thank you so much for your contribution

PR #102, opened 2026-08-06
fix(cdp): keep Playwright off the terminal with retained diagnostics

elijah-rou:
## Summary

- route Playwright output through a bounded, non-file pipe instead of the live terminal
- retain the latest 64 KiB of Node diagnostics through startup, runtime, and shutdown errors
- use a stable logger sink and enforce one managed Playwright lifecycle around playwright-go's package-global logger
- propagate browser, driver, and local-server shutdown failures to the caller

## Why

The Node driver can snapshot Bubble Tea's raw termios state and restore it after Bubble Tea returns the terminal to cooked mode. The shell is then left without normal echo, canonical input, signals, or output post-processing.

The first commit is Ian Swope's change from #99 with authorship preserved. This combined PR layers a reviewed follow-up on #99 because the original cross-fork branch is not writable. Maintainers can merge this PR, or merge #99 and cherry-pick the follow-up commit `3cd1b21`.

Refs #88

## Tests

- `go test -count=1 ./...`
- `go vet ./...`
- `go test -race -count=10 ./internal/player/cdp`
- `GOOS=darwin GOARCH=arm64 CGO_ENABLED=0 go test -c ./internal/player/cdp`

### DIFF
--- a/cmd/cdp_common.go
+++ b/cmd/cdp_common.go
@@ -4,6 +4,7 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"syscall"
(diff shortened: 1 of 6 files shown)


elijah-rou:
On second thought, I'd prefer to wait until #99 is merged rather than duplicate Ian's fix here. I'll keep the narrower logger-preservation and bounded startup-diagnostics follow-up as a draft until the base fix is settled.

PR #103, opened 2026-08-07
fix(cdp): preserve logs and retain startup diagnostics

elijah-rou:
## Status

Draft follow-up to #99. Do not merge before #99. This branch is stacked on Ian Swope's PR head, so GitHub will show that commit until #99 lands and this branch is rebased. The intended follow-up is commit `a1f648c` only.

## Summary

- avoid configuring Playwright output during the version-only driver check, which does not spawn a child
- pass one explicit Playwright logger so playwright-go does not redirect Go's process-wide standard logger
- keep the newest 64 KiB of piped Node output and attach it to Playwright install/startup errors
- preserve full error causes while keeping Node off the terminal

## Value

#99 fixes the terminal corruption. This follow-up addresses two separate consequences of routing output away from the terminal: playwright-go otherwise redirects the global standard logger to the supplied stderr writer, and discarding all Node output removes useful diagnostics when the driver fails to start.

## Tests

- `go test -count=1 ./...`
- `go vet ./...`
- `go test -race -count=10 ./internal/player/cdp`
- `GOOS=darwin GOARCH=arm64 CGO_ENABLED=0 go test -c ./internal/player/cdp`

### DIFF
--- a/internal/player/cdp/browser.go
+++ b/internal/player/cdp/browser.go
@@ -68,8 +68,6 @@ func linkHelper() {
 func isDriverUpToDate() bool {
 	driver, err := playwright.NewDriver(&playwright.RunOptions{
 		DriverDirectory: driverDir(),
-		Stdout:          driverOutput(),
-		Stderr:          driverOutput(),
 	})
 	if err != nil {
 		return false
@@ -103,13 +101,9 @@ func EnsureBrowser(onProgress func(string)) error {
 	// Also ensure the playwright driver is available (no browser install via playwright).
 	_ = os.Setenv("PLAYWRIGHT_SKIP_VALIDATE_HOST_REQUIREMENTS", "1")
 	onProgress("Fetching dependencies…")
-	if err := playwright.Install(&playwright.RunOptions{
-		DriverDirectory:     driverDir(),
-		SkipInstallBrowsers: true,
-		Stdout:              driverOutput(),
-		Stderr:              driverOutput(),
-	}); err != nil {
-		return fmt.Errorf("playwright driver: %w", err)
+	driverOptions, driverOutput := newDriverRunOptions(driverDir())
+	if err := playwright.Install(driverOptions); err != nil {
+		return fmt.Errorf("playwright driver: %w", addDriverOutput(err, driverOutput))
 	}
 
 	if chromeInstalled {
@@ -213,14 +207,10 @@ func extractDeb(debPath, destDir string) error {
 // runPlaywright starts the Playwright driver backed by our cached Chrome.
 func runPlaywright() (*playwright.Playwright, error) {
 	_ = os.Setenv("PLAYWRIGHT_SKIP_VALIDATE_HOST_REQUIREMENTS", "1")
-	pw, err := playwright.Run(&playwright.RunOptions{
-		DriverDirectory:     driverDir(),
-		SkipInstallBrowsers: true,
-		Stdout:              driverOutput(),
-		Stderr:              driverOutput(),
-	})
+	driverOptions, driverOutput := newDriverRunOptions(driverDir())
+	pw, err := playwright.Run(driverOptions)
 	if err != nil {
-		return nil, fmt.Errorf("playwright driver: %w", err)
+		return nil, fmt.Errorf("playwright driver: %w", addDriverOutput(err, driverOutput))
 	}
 	return pw, nil
 }
--- a/internal/player/cdp/browser.go
+++ b/internal/player/cdp/browser.go
@@ -68,6 +68,8 @@ func linkHelper() {
 func isDriverUpToDate() bool {
 	driver, err := playwright.NewDriver(&playwright.RunOptions{
 		DriverDirectory: driverDir(),
+		Stdout:          driverOutput(),
+		Stderr:          driverOutput(),
 	})
 	if err != nil {
 		return false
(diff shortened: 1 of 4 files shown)


ianswope:
Heads up that both this and my #99 were written before #94 landed, and it changed the shape underneath us.

`EnsureBrowser` in `browser.go` was split into `ensureBrowserARM64` and `ensureBrowserAMD64`, so there are now **two** `playwright.Install` call sites on Linux rather than one:

- `browser.go:245` — `ensureBrowserARM64`
- `browser.go:324` — `ensureBrowserAMD64`

Plus `runPlaywright` at `browser.go:432`, and the two in `browser_darwin.go` — five spots that spawn the driver.

This branch patches the single `EnsureBrowser` that existed at the time (`newDriverRunOptions` at line 105), so on rebase the hunk will apply to one of the two and silently leave the other still handing node a TTY. Easy to miss in review, since the diff looks complete either way. Same problem on mine.

Your `newDriverRunOptions` helper handles this better than my approach does: with one constructor, a missed call site shows up as a raw `&playwright.RunOptions{}` literal rather than an absent field. Might be worth a test asserting no `RunOptions` literal is constructed outside the helper, so the next split doesn't reintroduce it.

(`isDriverUpToDate` also builds a bare `RunOptions`, but only to read `driver.Version` — it never starts the driver process, so it isn't part of this.)

Also worth noting for anyone reviewing: retaining bounded output and appending it to the error is the right call. I went with `io.Discard`, which keeps the terminal clean but makes driver startup failures undebuggable — a bad trade that this avoids.

ianswope:
Heads-up: this is conflicting against current `main`, and I think my rebase of #99 is what stranded it — sorry about that.

The conflicts are in `browser.go` and `browser_darwin.go`, from #100 and #94, both of which touched those files after this branch's base (#91).

One thing worth knowing before you rebase: this branch carries a copy of my #99 commit (`e0582a2`), and #99's current head is `504dcae` — different content, not just a different hash, since I adjusted it for #100 and #94. A straight rebase would have you resolving conflicts in my outdated commit as well as your own.

Cleanest path is probably to let #99 land first (it's rebased onto current `main` and green), then rebase this onto `main` and drop `e0582a2`, leaving just `a1f648c` — the part that's actually yours. If you'd rather not wait, rebasing onto #99's current head (`504dcae`) gets you the same place.

No rush from me, and happy to do the rebase myself if that's easier — just say.

elijah-rou:
get #99 in first, this is add-on work

ianswope:
#99 landed as `4e18233`, so this is unblocked.

Note that `driverStderr()` and `driverLogger()` are both in `internal/player/cdp/driverio.go` on `main` now, so after a rebase your diff should reduce to the part that's yours: the bounded node-output buffer and attaching it to install/startup errors. The copy of my commit this branch was carrying should drop out at the same time.

ianswope:
Checking in on this one. Still happy to see it land if you have time. If you would rather hand it off, say so and I will pick it up with credit to you.

`main` has moved under it since August, mostly in your favour:

- The `playwright.Install` call your first hunk patches is no longer in `browser.go`. It moved into `installPlaywrightDriver` in the new `internal/player/cdp/driver.go` (function at `driver.go:37`, the call at `:42`), which the Linux and darwin paths now share. So the install-error half of this becomes one hunk instead of two.
- `runPlaywright` is unchanged in shape, now at `browser.go:458` and `browser_darwin.go:94`, so that half should rebase as-is.
- `driverStderr()` and `driverLogger()` have been in `driverio.go` since #99 landed, so the copy of my commit this branch was carrying should drop out on rebase, leaving just the bounded node-output buffer and attaching it to install and startup errors.

One heads-up: #129 landed yesterday and rewrote a fair amount of `browser.go`, but it left `runPlaywright` and the driver install alone, so it should not add conflicts for you. Everything above is against `a1587ca` (v0.8.0).

No rush from my side. I just did not want the branch to go stale silently.

elijah-rou:
@ianswope I am currently fairly swamped, I can maybe get to it towards the end of the week/weekend. I am happy for you to take it over if you wish.

ianswope:
Thanks, picked it up as #135.

It is not a rebase of this branch. #99 rewrote `driverio.go` after this was cut, so `driverOutput()` no longer exists and the diff here no longer applies. #135 re-applies your design to the current shape, with you as co-author on the commit. So the rebase I described in August as "drop my commit and you are done" was already out of date when I said it, which is worth saying plainly rather than leaving you to discover it.

What is kept as yours: the bounded 64 KiB buffer that is still a plain `io.Writer`, `addDriverOutput`, and the constructor. The constructor is the part that matters most, and #135 adds the test I suggested on this thread in August to go with it, scanning the package's non-test sources for a bare `playwright.RunOptions{` so the next split of `EnsureBrowser` cannot quietly reintroduce one.

One correction to this thread. Your note that playwright-go keeps the logger in package state was right, and so was the `driverLogger()` comment about `log.SetOutput`. They are different lines, `run.go:455` and `run.go:452`, and both matter. #135 cites each rather than picking one.

Worth knowing either way: the install path captures nothing today, because `d.log` is gated on `Verbose` and `SkipInstallBrowsers` skips `installBrowsers`, the only other writer. The behaviour change is all on the long-lived driver's stderr, which playwright-go wires to `cmd.Stderr` at `transport.go:113`.

@simonepelosi this one can be closed in favour of #135 whenever suits.

ianswope:
#135 is merged, so there is nothing left here to land. Worth closing when either of you gets a moment, since I cannot. Thanks again for the hand-off.

PR #104, opened 2026-08-07
fix(tui): keep status hints inside the border

ianswope:
## What's wrong

`padRight` returned over-long content unchanged (`if sw >= w { return s }`), so any
status line wider than the frame ran straight past the closing `│` and broke the
vertical rule. Both status builders also accepted a width argument and ignored it
(`_ int`).

The NORMAL nav line needs **140 columns** to fit on one row, so this fires on any
terminal narrower than that — which is how `v vibe` and `R radio` ended up sitting
outside the border. `go run . --demo` in a window under ~140 columns shows it.

## What changed

- **`padRight` now clips** with an ANSI-aware truncate. It's a global backstop for
  every row rather than the status bar alone, which also covers two pre-existing
  overflows I hit while testing: the `made with ❤️ by simonepelosi` footer (52
  columns in a 40-column box) and the vibe placeholder (70 in 60).
- **`wrapFit`** packs hints onto as many rows as they need, so a narrow terminal
  costs extra rows instead of dropped hints. Nothing is ever hidden.
- `statusNavContent`/`statusPlayContent` become `statusNavLines`/`statusPlayLines`
  returning `[]string`, with a `statusLines` combiner shared by the renderer and the
  height math.
- **`panelHeight`'s hardcoded overhead of 20 becomes `18 + len(statusLines)`.** This
  one isn't cosmetic: once hints wrap, the old constant under-counts and the frame
  overflows the terminal. At width 60 in a 40-row terminal it rendered 44 rows.

`charmbracelet/x/ansi` goes from indirect to direct in `go.mod` — it was already in
the module graph via lipgloss, so this adds nothing to the dependency tree.

## Tests

`TestView_HeightMatchesTerminal` and `TestView_StatusLinesRespectBorder` are the
regression guards. I checked both actually bite by reverting the `panelHeight`
accounting and the `padRight` clip in turn and watching them fail.

Full suite, `go vet`, and golangci-lint v2.11.4 all clean.

## Note on scope

There's a purely cosmetic change I deliberately kept out of this PR: tightening the
separator and shortening three labels takes the NORMAL line from 140 columns to 105,
so both rows stay single-row at ≥109. That's a UI-surface change and you've said you
prefer a minimal UI, so I've put it in #106 for you to accept or reject
separately.

**This fix stands on its own without it** — hints just wrap onto extra rows sooner on
a narrow terminal. The border never breaks either way.

### DIFF
--- a/internal/tui/model.go
+++ b/internal/tui/model.go
@@ -3052,9 +3052,10 @@ func (m *Model) renderBoxLayout() string {
 		sb.WriteString("├" + strings.Repeat("─", splitW) + "┴" + strings.Repeat("─", rightW) + "┤\n")
 	}
 
-	// ── Status bar (two lines: context/mode + playback) ──
-	sb.WriteString("│ " + padRight(m.statusNavContent(inner-2), inner-2) + " │\n")
-	sb.WriteString("│ " + padRight(m.statusPlayContent(inner-2), inner-2) + " │\n")
+	// ── Status bar (context/mode then playback, each wrapped as needed) ──
+	for _, line := range m.statusLines(inner - 2) {
+		sb.WriteString("│ " + padRight(line, inner-2) + " │\n")
+	}
 
 	// ── Bottom border ──
 	sb.WriteString("└" + strings.Repeat("─", inner) + "┘")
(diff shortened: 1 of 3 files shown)


simonepelosi:
Hey! Thank you for this, can you solve conflicts ?

ianswope:
Rebased onto current `main`.

The conflict was in `panelHeight()`. #90 made the now-playing section variable-height for art mode (`8 + nowPlayingHeight()`), and this PR makes the status bar variable-height (hints wrap instead of overflowing). Taking either side alone breaks the other, so both are measured now:

```go
fixedOverhead := 6 + m.nowPlayingHeight() + len(m.statusLines(m.width-4))
return max(3, m.height-fixedOverhead)
```

6 is the true fixed chrome: top, header, header divider, split divider, join divider, bottom.

The `go.mod` change dropped out entirely — `main` already promoted `github.com/charmbracelet/x/ansi` to a direct dependency, so this PR is now just the three TUI files.

Green on all three CI jobs.

simonepelosi:
Than you so much!

PR #105, opened 2026-08-07
build: add 'make install' and 'make uninstall'

ianswope:
## Why

`make build-with-token` leaves the binary in the working tree, so a copy already on
`PATH` can silently go stale behind a newer `./vibez`. Embedded developer tokens also
expire after 30 days, so that stale copy eventually starts failing against the catalog
with no obvious cause — it presents as an Apple problem rather than a build problem. I
lost an afternoon to exactly that.

## What this adds

- **`make install`** rebuilds with a freshly generated embedded token and copies the
  result to `$(BINDIR)`, defaulting to `$(HOME)/.local/bin`. Override with
  `make install PREFIX=/usr/local`, or set `BINDIR` directly.
- **`make uninstall`** removes the installed binary.

Credentials come from the environment or a gitignored `.env`, accepting either
`APPLE_PRIVATE_KEY` with the key contents or `APPLE_PRIVATE_KEY_FILE` with a path to
the `.p8` — the same inputs `refresh-token` already accepts.

README gets a short block under the existing build-from-source instructions.

## One thing worth your call

`install` duplicates about six lines of `.env` / `APPLE_PRIVATE_KEY_FILE` handling
from #91's `refresh-token` rather than factoring out a shared snippet. That was
deliberate — factoring it out means editing your merged target from inside an
unrelated PR. Happy to do it if you'd rather have one shared block; I just didn't want
to assume.

I also dropped my own `refresh-token` equivalent when #91 landed. Yours is better: it
adds `gen-devtoken -write` plus tests.

## Tests

Tested end-to-end with `PREFIX=$(mktemp -d)`: `.env` loaded, `APPLE_PRIVATE_KEY_FILE`
expanded, binary built and installed 755, `uninstall` removed it. Full suite,
`go vet`, and golangci-lint v2.11.4 clean.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,9 @@
 GOFLAGS=-tags webkit2gtk_4_1
 PKG_CONFIG_PATH=$(CURDIR)/pkg-config
+PREFIX ?= $(HOME)/.local
+BINDIR ?= $(PREFIX)/bin
 
-.PHONY: build run test lint clean release build-with-token refresh-token
+.PHONY: build run test lint clean release build-with-token refresh-token install uninstall
 
 build:
 	PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) go build -o vibez .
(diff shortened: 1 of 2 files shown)


simonepelosi (approved):
LGTM!

Issue #110, opened 2026-08-11
MPRIS: PropertiesChanged emits godbus's mutable stored value, racing with the next update


ianswope:
## Problem

Every `PropertiesChanged` signal vibez emits references the same map that the next property update writes into. Under `-race`, `go test -race -count=10 ./internal/player/mpris/...` fails with godbus's encoder reading the metadata map while a flush writes it.

This predates #101 — it reproduces on `main` (`d62f9f6`) in `TestMPRISPlayer_ReadState_*`. #101 is only the first change that trips it reliably, and it should not be held up for this.

## Mechanism

1. `prop.Export` allocates each property's storage once at registration and keeps a pointer to it (`prop/prop.go:220-222`).
2. `SetMust` → `p.set` → `dbus.Store([]any{v}, prop.Value)` merges the new value **into that allocation**. For maps, `storeMapIntoMap` calls `MakeMap` only when the destination is nil, otherwise doing in-place `SetMapIndex` (`dbus.go:180-195`).
3. `emitChange` passes `prop.Value` — the pointer — to `Emit` (`prop/prop.go:300`). `MakeVariant` does not dereference.
4. The dereference happens at **encode time**, in godbus's output goroutine (`encoder.go:151`), which takes no lock. `p.mut` protects `set` but not the encoder.

So `flush` (`internal/player/mpris/server.go:339-354`) queues a signal pointing at a map it is about to rewrite.

## Why the obvious fix doesn't work

Passing a freshly allocated map to `SetMust` changes nothing: `Store` copies its entries into godbus's existing allocation regardless. The aliasing is inside the `prop` package, not at our call site.

## Proposed fix

Stop letting the stored value be the emitted value: register the runtime-updated Player properties with `Emit: prop.EmitFalse` and emit `PropertiesChanged` from `flush` ourselves, built fresh each time.

That also fixes something orthogonal: `flush` currently calls `SetMust` 6–8 times per update, so a single track change emits 6–8 separate `PropertiesChanged` signals. The spec expects one batched signal per change, so this reduces bus traffic rather than adding any.

Happy to implement if the direction looks right — flagging it as an issue first since it changes how properties are exported rather than patching a line.

## Related

[...]

PR #111, opened 2026-08-11
fix(mpris): write every metadata key on each update

ianswope:
## Summary

- write the full MPRIS metadata key set on every update, with zero values where a field is absent

## Why

godbus merges each property update into the map it already holds rather than replacing it: `prop.Export` allocates the storage once at registration (`prop/prop.go:220-222`), and `dbus.Store` → `storeMapIntoMap` calls `MakeMap` only when the destination is nil, otherwise doing in-place `SetMapIndex` (`dbus.go:180-195`).

A key left out of an update therefore keeps its previous value instead of disappearing:

- a track with no artwork kept the previous track's `mpris:artUrl`, so clients showed the wrong cover
- a nil `Track` wrote only `mpris:trackid`, so stopping playback left the previous title, artist, album and length exported

Both are visible in any MPRIS client — playerctl, waybar, the GNOME/KDE media widgets.

## Scope

Rebased on `6b02d73`, so this sits on top of #101 and keeps its `trackPath`.

This deliberately does **not** address #110, which covers the related data race and the export change that would let keys be removed outright rather than blanked. A constant key set is the fix available without changing how properties are exported.

## On empty values rather than absent keys

MPRIS clients treat `""` and absent alike, and `xesam:album` already went out empty before this change, so this follows existing behaviour. `mpris:length: 0` when nothing is playing is the one deliberate judgement call — it replaces a stale non-zero length, so it is strictly better than what shipped, but it is a choice rather than an obvious default.

## Tests

- new `TestServer_Update_ClearsStaleMetadataKeys` and `TestServer_Update_ClearsMetadataWhenTrackGoesAway`; both fail on `main` without the change
- `go test -count=1 ./...`, `go vet ./...`, `golangci-lint` v2.11.4 (0 issues), `gofmt`


simonepelosi:
Nice catch on the `godbus` map-merging behavior (`storeMapIntoMap`)! Always populating the full set of 6 standard MPRIS keys with default/zero values cleanly prevents previous track metadata (like artwork or title) from lingering in media widgets and `playerctl`.

A few highlights from reviewing:

- **Zero Latency Impact**: Confirmed that `flush()` remains cleanly gated on `statusChanged`, `trackChanged`, or `seeked`, so regular position updates do not emit extra D-Bus messages.

I said that because initially I was sending a lot of D-Bus messages and the vibez UI was a little big laggy, so we should have in mind this infrastructure! 

LGTM! 🚀

Issue #113, opened 2026-08-12
Just installed i cannot login


ignotion:
It opens firefox and then it says :

"Error: Initialized with an expired token."

I've loged out, same result (something i could expect as i couldn't login)
I've tried to log via terminal (that opens the browser) and same result

ianswope:
That's enough to go on. The error names the cause precisely, and it isn't
anything on your end.

Release binaries embed an Apple Music developer token at build time, signed
for 30 days. v0.5.0 was published on 10 July, so its token expired around
9 August. MusicKit rejects it in `configure()`, before your Apple account is
involved at all. That's why logging out and back in makes no difference, and
why the terminal flow fails identically. There's nothing cached locally you
can clear.

Root cause and the lifecycle discussion are in #108. #65 was the same failure
at the previous expiry.

Two things worth knowing:

- There is no config workaround on a release build. Even with your own
  developer token, `ApplyEmbedded` (`internal/auth/devtoken.go:16`) overwrites
  it with the embedded one, so `apple_developer_token` in
  `~/.config/vibez/config.json` is ignored. A source build without an embedded
  token does honour it, but that needs an Apple Developer Program membership.
- A new release re-signs the token and unblocks everyone currently stuck.
  That's the realistic fix here.

One thing to confirm: does `vibez version` print 0.5.0, and did you install via
the install script or a release binary rather than building from source? If you
built it yourself the diagnosis is different.

ignotion:
Installed vía script, so I assume is exactly what you meant, it makes sense completely. Thanks for your help

cmdenis:
Just so that I understand: so does that mean that everything will be fixed on a future release?

ianswope:
@cmdenis Yes for the failure you're hitting, with a caveat about "everything".

A new release is built with a freshly signed developer token, so login works
again for anyone on that build. Nothing is needed from you, and there is no
config workaround on a release binary before then.

The caveat: Apple signs those tokens for 30 days, and a release doesn't change
that. v0.5.0 shipped on 10 July and stopped working around 9 August. The next
release starts the same clock, and #65 was the previous round of it. So a
release fixes this, then it recurs unless the lifecycle changes, which is what
#108 is for.

Two related fixes are open, but neither unblocks login on its own: #114 stops a
rejected developer token from deleting a working user session, and #118 covers a
token rejection on the search path. MusicKit needs a valid developer token
before it will authorize anyone, so both are about failing better rather than
logging in with an expired one.

[...]

Issue #108, opened 2026-08-09
Define a sustainable Apple developer-token rotation strategy


elijah-rou:
## Problem

Release builds embed an Apple Music developer JWT generated by `scripts/gen-devtoken`. The generator currently sets `exp` to 30 days after signing, so every installed copy of a release shares the same expiration date. MusicKit then rejects initialization with:

```
Initialized with an expired token.
```

That string comes from MusicKit's developer-token constructor, rather than its Music User Token handling. The existing expiry work in #28 and #71 handles the subscriber/user token, but cannot renew an expired developer credential. A 401 from the developer token can also be mistaken for an expired user session, causing vibez to clear the user token and enter a login flow that cannot initialize.

This timing is visible in the current release lifecycle: v0.5.0 was published on July 10- 2026, giving its embedded 30-day token a nominal expiration around August 9. Issue #65 was temporarily resolved by v0.3.0 shipping a newly signed token, but the same lifecycle can recur for every release.

PR #91 added `make refresh-token` for local builds, but release binaries still prefer the embedded token over a configured replacement. Allowing a newer configured token and extending generation to 180 days would mitigate the problem, but would still leave ordinary users broken after roughly six months if no newer release exists.

## Design decision needed

A static binary cannot renew this credential without either access to the MusicKit private signing key or a maintained source of newly signed tokens. Obviously the private key must not be distributed in the client, so the project needs to choose an explicit token lifecycle:

1. **Release-driven rotation:** publish a new version before the embedded token expires and rely on the self-updater. This is simple, but makes old releases time-limited and requires a release at least every six months.
2. **Remote token rotation:** retain an embedded bootstrap/fallback token while fetching and caching a newer signed token from a maintainer-operated endpoint or periodically updated artifact. This avoids forced application releases, but introduces availability, abuse/rate-limit, revocation, and trust considerations.
3. **User-provided credentials:** require users to generate their own developer token. This avoids central infrastructure but requires Apple Developer Program membership and substantially changes installation.
4. **Hybrid:** use remote rotation for release users, allow a newer configured token for advanced/local builds, and retain the embedded token as an offline fallback while valid.

Of course, there may be other ways. This is just a non-exhaustive list of what could be done.

Whichever approach is chosen should also:

- distinguish developer-token failure from Music User Token expiry so a valid user token is not cleared;
- attempt application or token updates before treating expiry as fatal;
- report the developer-token expiration and recovery path before opening MusicKit;
[...]

PR #114, opened 2026-08-12
fix(auth): keep the user token when the developer token is rejected

ianswope:
`ValidateToken` probed `/v1/me/storefront`, which requires both the developer
token and the Music User Token, and collapsed the answer to a bool. A 401 there
cannot say which credential Apple refused, but both callers treated it as an
expired session: `cmd/cdp_common.go` and `cmd/root_linux.go` printed "Session
expired - re-authenticating...", cleared `AppleUserToken` and saved the config.

So an expired developer token deleted a working session, and the re-login that
followed could not succeed either, since MusicKit needs a valid developer token
before it will authorize anyone. This is live right now: the token embedded in
v0.5.0 expired around 9 August, and #113 is someone hitting it.

`CheckTokens` returns a three-state result instead. When the combined request is
rejected, a second probe against `/v1/storefronts` sends the developer token
alone and attributes the failure:

- `UserTokenRejected` clears the token and re-authenticates, as before
- `DeveloperTokenRejected` reports the cause and leaves the session intact
- `TokensValid` proceeds

Anything the probes cannot establish, including a network failure on either
request, returns `TokensValid`. That keeps the old assume-valid behaviour, so a
flaky connection never costs someone a session.

This is only the attribution bug. It does not touch the lifecycle question in
#108, which is your call, and it does not change `ApplyEmbedded`.

Tested on linux/amd64: `go test ./...`, `go vet ./...`, golangci-lint v2.11.4
and gofmt all clean. `GOOS=darwin GOARCH=arm64 go build ./cmd/...` compiles.
`root_darwin.go` has no call site of its own and reaches this through
`cdp_common.go`, but it has not been run on macOS.

### DIFF
--- a/cmd/cdp_common.go
+++ b/cmd/cdp_common.go
@@ -45,10 +45,16 @@ func runCDPFlow(cfg *config.Config, opts tui.Options, onUserToken, onStorefront 
 
 		if cfg.AppleUserToken != "" {
 			prog.Send(tui.InitStatusMsg("Checking Apple Music session..."))
-			if !auth.ValidateToken(cfg.AppleDeveloperToken, cfg.AppleUserToken) {
+			switch auth.CheckTokens(cfg.AppleDeveloperToken, cfg.AppleUserToken) {
+			case auth.UserTokenRejected:
 				prog.Send(tui.InitStatusMsg("Session expired - re-authenticating..."))
 				cfg.AppleUserToken = ""
 				_ = cfg.Save("")
+			case auth.DeveloperTokenRejected:
+				// Re-authenticating cannot mint a working developer token, so
+				// keep the user token: it is still good once the build is.
+				prog.Send(tui.InitStatusMsg(auth.DeveloperTokenStatus))
+			case auth.TokensValid:
 			}
 		}
 
--- a/cmd/cdp_common.go
+++ b/cmd/cdp_common.go
@@ -43,6 +43,7 @@ func runCDPFlow(cfg *config.Config, opts tui.Options, onUserToken, onStorefront 
 			return
 		}
 
+		devTokenRejected := false
 		if cfg.AppleUserToken != "" {
 			prog.Send(tui.InitStatusMsg("Checking Apple Music session..."))
 			switch auth.CheckTokens(cfg.AppleDeveloperToken, cfg.AppleUserToken) {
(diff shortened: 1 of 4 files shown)


simonepelosi:
Thanks for working on this! Separating developer token failure from user token failure is a great fix for #108, and keeping the saved user session when a built-in dev token expires makes total sense.
I noticed a couple of minor edge cases worth addressing before we merge:
### 1. Handling non-2xx / non-401/403 responses in `probe()`
In [`internal/auth/validate.go`](https://github.com/simonepelosi/vibez/blob/main/internal/auth/validate.go):
```go
if resp.StatusCode == http.StatusUnauthorized || resp.StatusCode == http.StatusForbidden {
	return probeRejected
}
return probeAccepted
```

If /v1/me/storefront returns 401, but the tiebreaker request to /v1/storefronts hits a temporary server error (e.g. 500, 503, or 429 Rate Limit), probe() currently defaults to returning probeAccepted. This causes checkTokens() to conclude UserTokenRejected and wipe cfg.AppleUserToken.

To ensure server errors during the tiebreaker don't cost the user a session, we should explicitly check for 2xx status codes:

```go
if resp.StatusCode == http.StatusUnauthorized || resp.StatusCode == http.StatusForbidden {
	return probeRejected
}
if resp.StatusCode >= 200 && resp.StatusCode < 300 {
	return probeAccepted
}
return probeInconclusive
```

2. Status message overwrite in [cmd/cdp_common.go](https://github.com/simonepelosi/vibez/blob/main/cmd/cdp_common.go)
In cmd/cdp_common.go:

```go
case auth.DeveloperTokenRejected:
	prog.Send(tui.InitStatusMsg(auth.DeveloperTokenStatus))
...
status := hooks.initStatus
if status == "" {
	status = "Initializing vibez..."
}
prog.Send(tui.InitStatusMsg(status))
```

When DeveloperTokenRejected occurs, auth.DeveloperTokenStatus is sent to the TUI, but prog.Send(tui.InitStatusMsg("Initializing vibez...")) is called right below it, overwriting the warning message almost instantaneously. We should avoid sending "Initializing vibez..." if DeveloperTokenRejected was encountered.

Otherwise, the design and unit tests look super clean!

ianswope:
Both fixed, thanks.

1. `probe` now returns `probeAccepted` only for a 2xx. Anything else that is not a 401/403 is `probeInconclusive`, which `checkTokens` already routes to `TokensValid`, so a 429 or 5xx on the tiebreaker leaves the stored session alone. `TestCheckTokens_ServerErrorTiebreakerKeepsSession` covers 429/500/502/503 and fails on the old code with `UserTokenRejected`, which is the session wipe you described.

2. The generic progress line is skipped when the developer token was rejected, so the warning survives browser setup.

One thing that does not reach: `Starting audio engine...` still replaces the warning further down `runCDPFlow`, so on a fast start it is only briefly on screen. Making it stick would mean a persistent warning slot in the TUI rather than a status message. Happy to add that if you want it, but it seemed like more UI surface than this PR should carry.

simonepelosi:
Thank you so much for this quick commit! I really appreciate what you are doing! Sure, you're totally right, I'm fine wit hthe changes you made there! Please, move the changelog entry in the next release!

PR #115, opened 2026-08-18
fix(mpris): populate the test fake before it goes live on the bus

### LINKED ISSUE #107: Data race in MPRIS tests: fakeMPRIS fields mutated after the object is live on the bus
`go test -race` fails on `main` in the MPRIS package. It's a defect in the test
harness rather than in product code, but it means `-race` can't currently be used
on that package — including as evidence on incoming PRs.

## Reproduce

Needs a session bus (the tests skip without one):

```sh
go test -race -count=10 ./internal/player/mpris/
```

```
--- FAIL: TestMPRISPlayer_ReadState_WithMetadata (0.01s)
    testing.go:1712: race detected during execution of test
FAIL
```

## Cause

`fakeMPRIS` has a mutex and `Get`/`Set` take it correctly:

```go
func (f *fakeMPRIS) Get(iface, prop string) (dbus.Variant, *dbus.Error) {
	f.mu.Lock()
	defer f.mu.Unlock()
	...
```

But `newFakeMPRIS` requests the bus name and exports the object *before* returning,
so godbus's inbound worker goroutine can serve `Get` immediately — while the test
body then assigns those same fields with no lock held:

```
mpris_test.go:622    fake.playbackStatus = "Playing"
mpris_test.go:623    fake.volume = 0.9
mpris_test.go:624    fake.posUs = 30_000_000
mpris_test.go:646    fake.metadata = map[string]dbus.Variant{...}
```

The detector's trace is `fakeMPRIS.Get` (`mpris_test.go:374`) on godbus's
`(*Conn).inWorker` goroutine racing the write at `mpris_test.go:646`.

Two tests are affected — `TestMPRISPlayer_ReadState_Success` (three writes) and
`TestMPRISPlayer_ReadState_WithMetadata` (one). Only the second reports, because
it happens to lose the race first; both have the same defect.

## Suggested fix

Either give `fakeMPRIS` small setters that take `f.mu`, or let `newFakeMPRIS`
accept the initial field values so everything is populated before the object goes
live on the bus. The second is probably tidier — the tests only ever set these
once, immediately after construction.

## Context

Pre-existing on `main` and unrelated to any open PR. I noticed it while reviewing
#101, whose description cites `go test -race -count=10` as passing; that command
fails identically with and without those changes, so it isn't caused by them.
Filing separately so it doesn't get tangled up in that review.

ianswope:
Fixes #107.

`newFakeMPRIS` requested the bus name and exported the object before returning,
so godbus's inbound worker could serve `Get` while the test body was still
assigning `playbackStatus`, `volume`, `posUs` and `metadata` with no lock held.
`TestMPRISPlayer_ReadState_Success` and `TestMPRISPlayer_ReadState_WithMetadata`
both had the defect.

This takes the second option from the issue: `newFakeMPRIS` accepts seed options
that run before the export, so the guarded fields are never assigned once the
object can be read from the bus. The remaining writes all happen in the D-Bus
method handlers, which already take the mutex. Test-only change, no product code
touched.

Before, on `main` (6b02d73):

    $ go test -race -count=5 -run TestMPRISPlayer_ReadState ./internal/player/mpris/
    --- FAIL: TestMPRISPlayer_ReadState_Success (0.01s)

After:

    $ go test -race -count=10 -run TestMPRISPlayer_ReadState ./internal/player/mpris/
    ok  github.com/simone-vibes/vibez/internal/player/mpris  1.147s

Note that the whole-package command from the issue,
`go test -race -count=10 ./internal/player/mpris/`, still fails intermittently in
`TestServer_Update_DistinguishesCollidingTrackIDs`. That is #110, a separate race
in godbus's stored property value, not this one.

Full suite, `go vet` and golangci-lint v2.11.4 are clean.

### DIFF
--- a/internal/player/mpris/mpris_test.go
+++ b/internal/player/mpris/mpris_test.go
@@ -252,6 +252,11 @@ func TestParseMetadataMap_WrongTypeNocrash(t *testing.T) {
 
 // fakeMPRIS registers a minimal MPRIS2 object on the session bus.
 // It records method calls and returns configurable property values.
+//
+// Everything under mu is served to the bus by godbus's inbound worker goroutine
+// as soon as the object is exported, so tests must not assign these fields
+// directly.  Set the initial values through the options below, which run before
+// the export, and let the D-Bus method handlers do the rest under the lock.
 type fakeMPRIS struct {
 	mu             sync.Mutex
 	calls          []string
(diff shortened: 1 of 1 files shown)


simonepelosi:
Thanks for that <3

PR #116, opened 2026-08-18
fix(install): support macOS in the install script

ianswope:
## What

`scripts/install.sh` refuses to run on macOS, so the install one-liner in the README works for nobody on a Mac:

```
==> Checking platform…
    ✗  vibez currently supports Linux only (got: darwin)
```

Apple Silicon fails a second check right behind it — the blanket `if [ "${ARCH}" = "arm64" ]` rejection with its Linux-specific "no prebuilt Linux/arm64 binary" message — even though every release publishes `vibez_darwin_arm64.tar.gz` and `vibez_darwin_amd64.tar.gz` (both in v0.5.0, and `internal/updater/updater.go` already self-updates against exactly those names). macOS users are left with no binary install path at all, since the other documented route, `make build-with-token`, needs Apple Developer credentials.

## Why it needed two fixes

Opening the gate exposed a second break sitting behind it. macOS ships **its own** `/sbin/sha256sum` — `sha256sum (Darwin) 1.0`, options `[-bctwz]` — which wins the `command -v sha256sum` branch but rejects GNU's long options, so the `--check --status` pipeline failed and killed the install with a bogus message on a perfectly good archive:

```
==> Verifying checksum…
usage: sha256sum [-bctwz] [files ...]

    ✗  Checksum mismatch — the download may be corrupted. Please retry.
```

Note this is why the fix is not simply "let darwin through": the `shasum` fallback that was already in the script is unreachable on macOS, because the BSD `sha256sum` is found first.

## Changes

- **Platform gate** — a `case` on `$OS`: `linux` keeps its amd64-only rule and its existing build-from-source message (now parameterised, `Linux/${ARCH}`), `darwin` is accepted on both architectures, anything else is named in the refusal (`vibez supports Linux and macOS (got: freebsd)`).
- **Checksum verification** — read the expected digest from `checksums.txt` by exact filename (awk field match, tolerating the `*file` binary-mode form) and compare it to a digest taken with the bare `sha256sum` / `shasum -a 256` invocation, whose `<sha256>  <path>` output is identical on GNU coreutils, BSD and shasum. A missing checksum row is now a hard failure rather than printing `✓ Checksum OK` regardless, which the old skip branch did.
- **Desktop integration** — the XDG `.desktop`/icon step and its `update-desktop-database` / `gtk-update-icon-cache` refresh only run on Linux.
- **Chrome check on macOS** — an install ends by looking for Google Chrome in the same order `internal/player/cdp/browser_darwin.go` searches (`VIBEZ_CHROME_PATH` → `CHROME_PATH` → `/Applications` → `~/Applications`), warning with the `brew install --cask google-chrome` hint when it is absent, since vibez expects an installed Chrome on macOS and never downloads one.
- README: the one-liner is no longer labelled Linux-only, and states what it covers (Linux x86-64, macOS Apple Silicon + Intel) and that Linux/arm64 still builds from source. CHANGELOG entries under `[Unreleased] → Fixed`.

No Go code touched.

## Verification

Every run used an isolated `HOME`, so nothing landed in a real `~/.local/bin` or shell profile. Non-native platforms were faked with a `uname` shim on `PATH`.

| `uname -s` / `-m` | Result |
|---|---|
| Darwin / arm64 — **real hardware** (M-series, macOS 15) | `✓ macOS arm64` → `vibez_darwin_arm64.tar.gz` → `✓ Checksum OK` → installs → `vibez 0.5.0` runs (`Mach-O 64-bit executable arm64`); Chrome found; **no** XDG files created |
| Linux / x86_64 | unchanged: `✓ Linux x86_64` → `vibez_linux_amd64.tar.gz` → checksum OK → installs → desktop file + icon written |
| Linux / aarch64 | same build-from-source refusal as before, nothing written |
| FreeBSD / x86_64 | `✗ vibez supports Linux and macOS (got: freebsd)` |
| Darwin / ppc64 | `✗ Unsupported architecture: ppc64` |

Also: `sh -n`, `bash -n`, `zsh -n` clean; `go build ./...`, `go vet ./...`, `go test ./...` pass on darwin/arm64 (16 packages).

One gap, stated plainly: the GNU-coreutils side of the checksum change is reasoned from the bare-invocation output format being identical, not executed — no Linux box was available here. The Linux/x86_64 run above exercised the new code path, but with macOS's BSD `sha256sum`.

Refs #11.

## Out of scope

Two things I found but deliberately left alone:

- The PATH block appends `export PATH="${HOME}/.local/bin:${PATH}"` even when `VIBEZ_INSTALL_DIR` points somewhere else, so a custom install dir never reaches `PATH`. Pre-existing on Linux too; happy to send a separate PR.
- macOS release binaries are built `CGO_ENABLED=0` on an ubuntu runner (`.goreleaser.yml`). Worth settling before #61's CoreAudio player lands, since a cgo darwin backend can never ship from that job — separate discussion.


simonepelosi:
Thank you so much for this!

PR #118, opened 2026-08-19
fix(apple): fall back to catalog search when amp-api rejects the token

ianswope:
Refs #93, and the other half of #98. That one made the failure visible; this one makes it recoverable.

## The problem

The catalog-songs leg of `Search` goes through `amp-api.music.apple.com` (apple.go:23, request built at :415), because it is the endpoint that returns `extendedAssetUrls` and therefore the only one that can tell us whether a result is actually streamable.

It is also an undocumented web-player endpoint, and it can reject a developer/user token pair that the supported endpoints accept. That is what #93 turns out to be: 401 on every catalog-song request while the library and album/playlist legs keep working. After #98 that surfaces as a warning instead of silence, but the songs leg still returns nothing, so discovery has nothing to refill from and the user-visible symptom is unchanged.

## What this changes

When the amp-api request fails, the leg retries against Apple's supported `/catalog/{storefront}/search?types=songs` endpoint rather than giving up.

That endpoint does not return `extendedAssetUrls`, so `hasStream()` would discard every result it produces, for lacking a field the endpoint never sends. The filter is therefore gated on `verifiedStreams`, set only when the amp-api response succeeded. Fallback results are returned unfiltered, and playback stays the final authority on storefront availability.

This is the `extendedAssetUrls` decision I flagged in #98 as yours. I have taken the narrow reading: relax the filter only on the path that cannot populate the field, and never on the amp-api path. If you would rather the fallback did not exist, or would rather it return nothing than return unverified results, say so and I will cut it back.

## Tests

- `TestSearch_CatalogSongsFallBackWhenAmpAPIRejects`: amp-api answers 401, the fallback is called exactly once, and a song with no `extendedAssetUrls` survives into the results with no warning raised.
- `TestSearch_StreamFilterStillAppliesWhenAmpAPISucceeds`: amp-api succeeds, the fallback is never called, and a purchase-only result is still dropped.

`go build ./...`, `go vet ./...`, `go test ./...`, `gofmt -l` and `golangci-lint run` (v2.11.4) are all clean on linux/amd64.

### DIFF
--- a/internal/provider/apple/apple.go
+++ b/internal/provider/apple/apple.go
@@ -367,8 +367,9 @@ func (a *AppleProvider) Search(ctx context.Context, query string) (*provider.Sea
 		err       error
 	}
 	type catSongsOut struct {
-		songs []songResource
-		err   error
+		songs           []songResource
+		verifiedStreams bool
+		err             error
 	}
 	type catCollOut struct {
 		albums    []albumResource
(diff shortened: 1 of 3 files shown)


internal/provider/apple/apple.go
```diff
@@ -419,8 +420,25 @@ func (a *AppleProvider) Search(ctx context.Context, query string) (*provider.Sea
...
+		if err := a.do(req, &resp); err == nil {
+			catSongsCh <- catSongsOut{songs: resp.Results.Songs.Data, verifiedStreams: true}
+			return
+		}
+
+		// amp-api is an undocumented web-player endpoint and can reject otherwise
+		// valid developer/user tokens. Fall back to Apple's supported catalog
+		// search. It omits extendedAssetUrls, so playback remains the final
+		// authority for storefront availability.
+		fallbackEP := fmt.Sprintf("/catalog/%s/search?term=%s&types=songs&limit=25",
```
simonepelosi: Can we avoid to have a fixed limit there ?

simonepelosi:
Thank you so much! I reviewed it and I left a very small comment! <3

ianswope:
Rebased onto main to clear the CHANGELOG conflict from #111 and #116. The two code commits are unchanged, only the changelog line moved.

I cannot see the comment though, and nothing shows on the PR from my side. I think the review is still pending submission.

simonepelosi:
Thank you so much! Since I already released the 0.6.0, can you move the changelog entry in the next block? We will release the 0.6.1 with that improvement <3

Issue #65, opened 2026-07-01
Potential expiry of token


StarlightAbove:
Hi! I tried to log back into vibez after an update and it seems to be rejecting my login with "Error: Initialized with an expired token." I am unsure of how to deal with this, I presume it either has to do with how my OS has cached the key or if it is a key-thing over on your side.

simonepelosi:
Hey! Did you tried to do something like `vibez auth login` ?

StarlightAbove:
Yep. No dice.

simonepelosi:
Ok I may have found and fixed the issue, I will release a new version soon!

simonepelosi:
Release 0.3.0 out now !

simonepelosi:
Please, let me know if I solved the issue <3

StarlightAbove:
Solved! Closing this issue now.

PR #119, opened 2026-08-21
build(devtoken): raise the developer token TTL to Apple's ceiling

ianswope:
Refs #108. Item 1 of the plan you settled on there: a 180 day developer token, nothing else.

## What changes

`scripts/gen-devtoken` signed tokens for 30 days. This raises that to Apple's ceiling, 15777000 seconds, behind a named `tokenTTL` constant so the value has somewhere to be explained.

Releases embed the token at build time, so the old TTL meant a binary stopped reaching the catalog long before it stopped being the newest release. #113 and #65 are both that, one expiry apart.

Two things this does not do:

- It does not help any binary already published. v0.6.1 carries a 30 day token signed on 20 August, so it lapses around 19 September regardless. A release cut after this lands is what actually puts a six month token in front of users.
- It widens the window in which a leaked token stays usable, which is the cost I flagged in the issue. The trade still favours it: extraction difficulty is unchanged, and the alternative is a guaranteed monthly breakage.

Items 3 and 4 stay separate. 4 is already in via #109. I will send the `DeveloperTokenRejected` update prompt as its own PR.

## Tests

Nothing asserted the production expiry before this. Both existing fixtures rebuild the claims by hand rather than calling `generateToken`, so `exp` was the one claim no test covered, and a drift past Apple's limit would have surfaced at token exchange rather than at build time.

`TestGenerateTokenExpiry` signs a token with the real code path and asserts `exp` lands inside `[before + tokenTTL, after + tokenTTL]`, and fails outright if `tokenTTL` ever exceeds the six month ceiling. The two hand built fixtures now reference `tokenTTL` instead of restating 30 days, so they cannot disagree with the generator again.

`go build ./...`, `go vet ./...`, `go test ./...`, `gofmt -l` and `golangci-lint run` (v2.11.4) are all clean on linux/amd64.

### DIFF
--- a/scripts/gen-devtoken/main.go
+++ b/scripts/gen-devtoken/main.go
@@ -3,8 +3,7 @@
 // signed token to stdout so CI can capture and inject it at build time.
 //
 // With -write it instead writes the token into apple_developer_token in the
-// vibez config file, which is handy for local (non-embedded) builds where the
-// token expires every 30 days.
+// vibez config file, which is handy for local (non-embedded) builds.
 //
 // Required env vars:
 //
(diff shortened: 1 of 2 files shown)


simonepelosi (approved):

simonepelosi:
Thank you so much for that ❤️

Issue #117, opened 2026-08-18
Darwin release builds are CGO_ENABLED=0 on a Linux runner, so native macOS backends can't ship


ianswope:
### What

`.goreleaser.yml` builds the darwin artefacts with `CGO_ENABLED=0` on the `ubuntu-latest` release runner:

```yaml
  - id: vibez-darwin
    env:
      - CGO_ENABLED=0
      - GOTOOLCHAIN=local
    goos:
      - darwin
    goarch:
      - amd64
      - arm64
```

That is fine for what ships today — I checked the published `vibez_darwin_arm64.tar.gz` from v0.5.0 on Apple Silicon and it is ad-hoc linker-signed (`flags=0x20002(adhoc,linker-signed)`) and runs — so this is not a bug report about the current release.

It becomes load-bearing the moment a native macOS backend lands.

### Why it matters now

The local-tracks player (#61, PR #62) implements macOS playback as a cgo AudioToolbox backend, with a `player_darwin_nocgo.go` stub for builds without cgo. With the release job as it stands, every published mac binary takes the stub path: the CoreAudio player would exist in the source tree, pass CI, work for anyone running `make build` — and simply not be there for anyone who installed from a release or via the install script. Silently, with no error at build time.

The same ceiling applies to anything else native on macOS later (a Now Playing / `MPNowPlayingInfoCenter` bridge for media keys is the obvious candidate, since Linux gets that through the MPRIS server and macOS currently gets nothing).

### Options

1. **Build darwin on a macOS runner.** GoReleaser supports splitting the build across runners and merging the release (`goreleaser release --split` plus a merge job), or the release workflow can grow a separate `macos-latest` job whose artefacts are attached to the same release. CI already builds and tests on `macos-latest`, so the runner is a known quantity. Cross-compiling amd64 from an arm64 mac stays possible with `CC="clang -arch x86_64"`.
2. **Accept the stub on macOS and document it** — release binaries stream via Chrome/CDP only, and local-file playback on macOS requires building from source. Cheapest, but it means the feature's headline platform support is source-only.

Option 1 is the one that matches what #62 is promising users, but it touches release infra and the garble/token step, so it is your call rather than something to land quietly alongside a feature PR.

### Ask

A decision before #62 merges, so the answer is not discovered by a macOS user who installed a release and found local playback missing. Happy to prototype whichever direction you prefer.

simonepelosi:
[...]

Issue #96, opened 2026-07-19
seamless transitions


vmax1g:
is there a possibility that seamless transitions will be added in the future?

ianswope:
@vmax1g Not as things stand, and the reason is structural rather than a missing setting, so it's probably worth explaining where the gap comes from.

Playback runs through MusicKit.js in a headless Chrome page, and vibez keeps the queue on its own side (`_q` in `internal/player/web/musickit.html`) rather than handing it to MusicKit. It plays exactly one song at a time: on every track change `_doPlayAt` waits for the player to settle, calls `m.setQueue({songs:[id]})` with a single item, then plays it (musickit.html:373-410). Auto-advance is manual as well, with a `playbackStateDidChange` listener watching for `completed` before starting the next index (musickit.html:471-482).

So every boundary is a full teardown and rebuild: track ends, completed fires, new one-item queue, DRM and storefront resolution, buffer, play. Nothing prebuffers the next song, so a pause between tracks is unavoidable in the current design.

The way out would be to hand MusicKit the whole queue once and navigate with `skipToNextItem`, which lets it prebuffer. That design is actually already described in a comment at musickit.html:216-227, but the code doesn't do it yet: `skipToNextItem` and `queue.append` appear only in that comment. The same comment records why it's awkward, in that `setQueue` mid-session gets silently ignored with `CONTENT_EQUIVALENT` and desyncs vibez's queue from MusicKit's. The per-track library fallback for unavailable catalog songs also assumes one-song-at-a-time and would need rethinking.

One caveat on expectations: MusicKit.js doesn't expose true sample-accurate gapless playback, so realistically this means a much shorter gap rather than none at all.

Whether it's worth doing is @simonepelosi's call. It's a decent chunk of work in the most delicate part of the player.

vmax1g:
@ianswope to clarify what im actually after, not true crossfading, just for albums that are mastered with no gap between tracks (like they would play gapless on apple music's own app/web player), would it be possible for vibez to not add its own extra pause on top? sounds like that's basically what the setQueue + skipToNextItem approach you described would fix.

ianswope:
Your reading is right. That pause comes from vibez, not from anything Apple's player does on the same album.

[...]

PR #120, opened 2026-08-22
build(release): build darwin on a macOS runner with CGo enabled

ianswope:
Refs #117. Option 1, the one you picked: darwin builds on `macos-latest` with `CGO_ENABLED=1`.

## What changes

`.goreleaser.yml` flips the darwin build to `CGO_ENABLED=1`, and `release.yml` grows a
`build-darwin` job on `macos-latest` that produces the two darwin archives and hands them to
the existing ubuntu job, which attaches them to the same release through
`release.extra_files` and covers them in `checksums.txt` through `checksum.extra_files`. One
release, one checksum file, same artefact names as today.

Both builds carry a templated `skip`, so a single config serves both runners: the mac job
sets `VIBEZ_SKIP_LINUX`, the linux job sets `VIBEZ_SKIP_DARWIN`, and with neither set the
config behaves exactly as it does now.

**A correction to my own issue text:** I offered `goreleaser release --split` plus
`continue --merge` as the mechanism. That is Pro-only, as is the `prebuilt` builder, which
would have been the other clean way to pass binaries between runners. Hence the artefact
handoff.

Three details worth knowing:

- `macos-latest` is arm64, so darwin/amd64 is a cgo cross-compile and needs
  `CC=clang -arch x86_64`. That `overrides` env merges into the build env rather than
  replacing it, so `CGO_ENABLED=1` still reaches the amd64 target — measured, not assumed.
- The developer token is signed once per job. A masked value cannot be handed between jobs,
  and two tokens signed from the same key are equivalent.
- `extra_files` globs quietly when nothing matches. A lost artefact would publish a release
  with no mac binaries and no error anywhere, so the ubuntu job checks both archives exist
  and contain a `vibez` binary before goreleaser runs.

## Reproduced on Apple Silicon

go1.26.1 darwin/arm64, no local edits beyond the config in this PR.

Against `main` the flip is inert, which is worth stating plainly: every cgo package in the
tree is `//go:build linux`, so nothing about today's feature set changes. To make cgo
observable I ran the same config against #62's head `a34eae3`:

| check | result |
|---|---|
| `garble -literals -tiny` over the AudioToolbox package | builds, both slices |
| `otool -L` on darwin/arm64 | links `/System/Library/Frameworks/AudioToolbox` |
| `otool -L` on darwin/amd64 (cross-compiled) | links `/System/Library/Frameworks/AudioToolbox` |
| same tree at `CGO_ENABLED=0` | no AudioToolbox — the stub path this removes from releases |
| binaries run | arm64 natively, amd64 under Rosetta, both report the injected version |
| garble applied | 0 readable module-path strings, matching the published v0.6.1 binary |
| archive shape | `CHANGELOG.md`, `LICENSE`, `README.md`, `vibez` — same as published |
| `checksum.extra_files` digests | match `shasum -a 256` |

The stub is the concrete failure this prevents: `player_darwin_nocgo.go`'s `New()` returns
`local playback requires CGo on macOS — build with CGO_ENABLED=1`, and that is what every
published mac binary would have done with #62 merged and the release job unchanged.

## Dry run of the whole flow

The two steps this machine cannot exercise are the cross-job artefact handoff and the
`release.extra_files` upload, so I ran the whole flow on a throwaway tag in my fork, on a
branch with #62 merged in so the outcome was observable end to end
([run](https://github.com/ianswope/vibez/actions/runs/32539261397)):

- `build-darwin` on `macos-latest`: linux skipped, `darwin_arm64_v8.0` and `darwin_amd64_v1`
  built and archived, 2m20s.
- The merged env for `darwin_amd64_v1` came out as
  `[CC=clang -arch x86_64 CXX=clang++ -arch x86_64 CGO_ENABLED=1 GOTOOLCHAIN=local]`.
- `release` on `ubuntu-latest`: both archives arrived in `dist-darwin`, darwin skipped,
  `linux_amd64_v1` built, four assets published — `vibez_linux_amd64.tar.gz`,
  `vibez_darwin_amd64.tar.gz`, `vibez_darwin_arm64.tar.gz`, `checksums.txt`.
- `checksums.txt` covers all three tarballs, and the darwin arm64 digest matches
  `shasum -a 256` run against the asset downloaded from the release.
- That published `vibez_darwin_arm64.tar.gz`, unpacked and run on this Mac: Mach-O arm64,
  links `AudioToolbox`, garbled, ad-hoc linker-signed, prints its version.

The scaffolding for that run was a stubbed token step, since my fork has no `APPLE_*`
secrets, and `release.github.owner` pointed at the fork with `draft: true`. None of it is in
this PR, and the tag and draft release are deleted.

## Two things I did not touch

`goreleaser check` fails on `main` today: `builds.gobinary` is deprecated in favour of
`builds.tool`, on both builds. Pre-existing and unrelated to this, so I left it alone. It can
go out as its own change if you want it cleared.

Signing and notarization are out of scope here. The arm64 slice comes out ad-hoc
linker-signed exactly as the current release does, and the amd64 slice is unsigned exactly
as the current release is, so nothing regresses; a Developer ID identity would be a
different conversation.

Linux is untouched, including the arm64 gap in the comment there.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,7 +12,64 @@ env:
   FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
 
 jobs:
+  # Darwin is built here rather than in the release job below because its
+  # native backends are CGO: a macOS runner is the only place clang and the
+  # macOS SDK are available. The archives are handed to the release job, which
+  # attaches them to the same release via extra_files.
+  darwin:
+    name: build-darwin
+    runs-on: macos-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.26'
+          cache: true
+
+      # Signed here as well as in the release job: a masked value cannot be
+      # handed between jobs, and two tokens signed from the same key are
+      # equivalent. garble installs via the goreleaser before-hook.
+      - name: Generate Apple Developer Token
+        id: devtoken
+        env:
+          APPLE_KEY_ID: ${{ secrets.APPLE_KEY_ID }}
+          APPLE_TEAM_ID: ${{ secrets.APPLE_TEAM_ID }}
+          APPLE_PRIVATE_KEY: ${{ secrets.APPLE_PRIVATE_KEY }}
+        run: |
+          TOKEN=$(go run ./scripts/gen-devtoken)
+          # Mask the token so it never appears in logs
+          echo "::add-mask::$TOKEN"
+          echo "token=$TOKEN" >> "$GITHUB_OUTPUT"
+
+      - name: Build darwin archives
+        uses: goreleaser/goreleaser-action@v6
+        with:
+          version: '~> v2'
+          args: release --clean --skip=publish
+        env:
+          APPLE_DEV_TOKEN: ${{ steps.devtoken.outputs.token }}
+          VIBEZ_SKIP_LINUX: '1'
+          # Prevent Go from downloading a different toolchain into GOMODCACHE,
+          # which garble cannot patch (overlay restrictions on GOMODCACHE).
+          GOTOOLCHAIN: local
+
+      - name: Upload darwin archives
+        uses: actions/upload-artifact@v4
+        with:
+          name: darwin-archives
+          path: dist/*.tar.gz
+          if-no-files-found: error
+          # Consumed by the next job; the same bytes ship on the release.
+          retention-days: 1
+
   release:
+    name: release
+    needs: darwin
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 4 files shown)


ianswope:
@simonepelosi #120 has been green since I opened it and I don't think it has had a look yet. Flagging that one over the rest because of what sits behind it: you wanted it in before #62, and on #96 the local-files half of Phase 2 stays invisible on macOS until release binaries are built with cgo.

No rush if you are mid-something. The part worth the most scrutiny is the artefact handoff, since `extra_files` globs quietly when nothing matches, which is why the ubuntu job asserts both archives exist and contain a binary before goreleaser runs. Happy to walk through it.

simonepelosi:
Thanks for the ping and for putting this together, @ianswope!

The artifact handoff and dual-runner setup look very clean. Having the explicit Check darwin archives step in release.yml to prevent silent omissions with extra_files is great. The token generation/masking per job and the cross-compilation flags (CC=clang -arch x86_64) for Apple Silicon runners are solid.

LGTM! There is just a minor conflict in CHANGELOG.md against latest main (from #121). I solved it!

PR #121, opened 2026-08-22
fix(auth): act on a rejected developer token instead of only reporting it

ianswope:
Refs #108, item 3 of the plan there, and the piece #113 actually needs.

## The problem

#114 made a rejected developer token survivable: the session is kept and the status bar reads
`Developer token rejected - update vibez to continue`. The advice is correct, and on a lot of
runs the user cannot act on it.

`CheckAndUpdate` looks at most once every 24 hours, gated by a stamp in the cache dir. Anyone
who started vibez yesterday, and whose embedded token expires today, is told to update by the
same run that skipped the check that would have updated them. The WebKit fallback path is
worse: it runs no update check at all, so there the message has never had anything behind it.

And when a token expires *before* a newer release exists — #113 exactly — "update vibez to
continue" is a dead end no matter what the throttle does.

## What changes

`updater.UpdateNow` asks GitHub straight away, ignoring the daily window, and reports what came
back as an outcome rather than just a path. The rejection branch calls it, installs and re-execs
when it can — reusing the restart plumbing already there for the startup check — and otherwise
names the actual situation:

| situation | what the user gets |
|---|---|
| newer release, updates allowed | installs it and restarts |
| already the newest release | `… - this is already the newest release, so a newer build has to be published first` |
| `--no-update` in force | `… - v0.6.1 is available; restart without --no-update to install it` |
| install cannot replace itself | `… - v0.6.1 is available, but this install cannot replace itself; reinstall to update` |
| the check itself failed | unchanged: `… - update vibez to continue` |

The second row is the #113 case. A failed check learns nothing, so it keeps the existing
wording rather than guessing.

`--no-update` is honoured: it reports the release and installs nothing. The asset lookup runs
*before* that gate on purpose, so a platform with no published archive — linux/arm64 today —
is never told to drop `--no-update` for a build that is not there. There is a test pinning
that ordering.

`CheckAndUpdate` keeps its signature and its throttle. Both it and `UpdateNow` now go through
one core that takes the endpoint, which is what makes the outcomes testable, and
`os.Executable` is indirected for the same reason, so the install test replaces a temp file
instead of the test binary.

## Verified

On Apple Silicon against the live Apple and GitHub endpoints, with `HOME` pointed at a temp dir
so the real update stamp stayed untouched:

| case | result |
|---|---|
| garbage credentials | `DeveloperTokenRejected` |
| `current=99.0.0` | newest-release message, no download |
| `current=0.1.0` + `--no-update` | names v0.6.1, no download |
| `current=0.1.0`, updates allowed | downloads v0.6.1, verifies the checksum, installs, returns the path to re-exec |

Nine unit tests cover the outcomes against an `httptest` release endpoint, including the
ordering rule above and a checksum mismatch leaving the binary in place. `go vet`,
`go test ./...`, `gofmt` and `golangci-lint` are clean here on darwin/arm64.

One gap worth stating: `cmd/root_linux.go` is the only file I cannot compile on this machine,
since cross-compiling the webkit/gstreamer cgo target needs a Linux toolchain — that is
pre-existing, `GOOS=linux go build ./...` fails the same way on `main`. `build-vet-test-linux`
is its compile check and I will confirm it green before asking you to look.

## Notes

Items 1 and 4 of #108 are already out as #119 and #109, so this leaves only item 2, the one we
agreed against.

The CHANGELOG entry lands in the same `## [Unreleased]` block as #120's, so whichever of the
two merges second may want a one-line merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- **A rejected developer token now acts on the update instead of only naming it** — the startup update check looks at most once every 24 hours, so a build whose embedded token had just expired could print `Developer token rejected - update vibez to continue` on the same run that skipped the check which would have fixed it. That path now asks GitHub immediately, installs and restarts when a newer release exists, and otherwise says which case it is: nothing newer published yet, `--no-update` in force, or an install that cannot replace itself. The WebKit fallback path, which ran no update check at all, gets the same handling. Refs #108, #113.
+
 ## [0.6.1] — 2026-08-21
 
 ### Fixed
--- a/cmd/cdp_common.go
+++ b/cmd/cdp_common.go
@@ -55,7 +55,23 @@ func runCDPFlow(cfg *config.Config, opts tui.Options, onUserToken, onStorefront 
 				// Re-authenticating cannot mint a working developer token, so
 				// keep the user token: it is still good once the build is.
 				devTokenRejected = true
-				prog.Send(tui.InitStatusMsg(auth.DeveloperTokenStatus))
+				// The update check above may have been inside its once-a-day
+				// window and returned without asking, which is exactly what
+				// would leave the user reading "update vibez" on a build that
+				// could have updated itself. Ask again, and say what came back.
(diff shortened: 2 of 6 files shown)


simonepelosi:
Looks fantastic @ianswope! Great tests and very clean handling of both CDP and WebKit restart paths. LGTM 🚀

Issue #122, opened 2026-08-28
Bug: when selecting a single time from Feed an Album, it gets inserted in the queue multiple times.


GiuseppeCucinotta:
<img width="917" height="621" alt="Image" src="https://github.com/user-attachments/assets/8624429c-6a97-4a17-8692-09140fd60ec1" />

simonepelosi:
Hey! Thank you for reaching out! Can you give me the steps to reproduce the issue? @GiuseppeCucinotta

GiuseppeCucinotta:
> Hey! Thank you for reaching out! Can you give me the steps to reproduce the issue? [@GiuseppeCucinotta](https://github.com/GiuseppeCucinotta)

It doesn't always bugs like this, but when it does the procedure is: from the "feed" screen, when I select an album that I have in that section, every once in a while it adds that album multiple times in the queue, simple as that. 
Tho is worth noting that there's multiple "now playing" indicators (as you can see in the photo). It seems like more a visual bug then the album that is put in the queue multiple times.

ianswope:
Reproduced both halves. This is two separate bugs that happen to show up in the same screenshot.

**The duplicate rows are real.** 91 tracks is the 13-track album seven times over.

The feed has three actions and only two of them can stack:

- `enter` replaces the queue, so pressing it repeatedly still leaves 13 tracks.
- `tab` (queue) and `shift+tab` (play next) append, and nothing tracks a fetch that is already in flight. A second press while the first is still running appends a second full copy.

`enter` also sets a loading flag, so the panel visibly reacts. `tab` sets nothing, and the album fetch runs under a 20s timeout, so there is no feedback at all while it waits. Press again, or hold the key long enough for terminal auto-repeat, and each repeat stacks another copy. That fits the "every once in a while" part.

@GiuseppeCucinotta which key are you using to add the album, `enter` or `tab`? That tells us whether this is the whole story.

**The repeated now-playing markers are a second bug, and the duplication is not what causes them.** `queuePanelLines` picks the marker with `t.Title == currentTitle` (`internal/tui/model.go:3392`), so every row whose title matches the playing track gets one. Two different songs that share a title are enough to mark both rows with nothing duplicated at all.

The two fixes are independent. I will send the marker one first since it is contained. For the duplication, the guard belongs on the fetch rather than the key: refuse a feed fetch for an item that already has one outstanding. Happy to take that too if nobody has started.

GiuseppeCucinotta:
[...]

PR #123, opened 2026-08-31
fix(tui): mark the playing queue row by identity, not by title

ianswope:
Fixes half of #122.

The queue panel picked the now-playing marker with a title comparison (`t.Title == currentTitle`), so every row whose title matched the playing track got one. The screenshot on #122 shows a queue that had picked up seven copies of the same album and marked all seven, but duplication is not required to see it: two different songs that share a name mark both rows.

`playingQueueIdx` now locates the playing track by playback ID and takes the first hit, matching how `dropQueueAfter` finds its seed. Title matching is kept as a fallback rather than removed, because the `CONTENT_UNAVAILABLE` library fallback can swap a catalog entry for its library copy mid-playback and leave the two IDs legitimately different. Either way only one row is ever marked.

The other half of #122, the album being appended to the queue several times, is a separate bug in the feed's append path and is not touched here.

### Tests

Four cases in `TestQueuePanelLines_*`. The two that target the bug fail on `main` (3 markers and 2 markers respectively) and pass here. The other two are regression guards: one for the library-fallback path where the IDs diverge, one for a playing track that is not in the queue at all.

Gate run on linux/amd64: `go test ./...`, `go vet ./...` and `gofmt -l .` are clean, and `golangci-lint` reports the same two pre-existing findings on this branch as on `main`, with none added.


simonepelosi:
Thanks @ianswope! Clean fix for #122. Matching on `PlaybackID` and capping to a single row prevents duplicate markers, and keeping title matching as a secondary pass is a great safeguard for mid-stream library swaps.

Tests cover all the edge cases nicely. Just needs a quick rebase on `main` to catch up with #131.

LGTM, ready to merge once rebased!

PR #124, opened 2026-08-31
feat(player): let MusicKit advance the queue instead of rebuilding it per track

ianswope:
Phase 1 of the plan in #96.

vibez rebuilt MusicKit's queue with `setQueue({songs:[one id]})` at every track boundary, so each transition paid a round trip that landed exactly where the gap is audible. This builds the queue once with `setQueue({items:[...]})` and lets MusicKit advance it. `items:` takes resolved descriptors of either kind, so a library song no longer forces the whole queue back to per-track rebuilds.

### What changed

- `musickit.html` builds the queue up front and hands advancement to MusicKit. `next`/`prev` delegate to `skipToNextItem`/`skipToPreviousItem` while it holds authority, and fall back to `_playAt` once the native queue is exhausted.
- Insertions go through `playNext`/`playLater`, which call `deferPlayback()` internally. That is the sanctioned way to extend a live queue without restarting playback.
- Library copies are warmed into `_libItems` when the queue is built and when `playLater` extends it, so the `CONTENT_UNAVAILABLE` fallback can swap from memory rather than paying a round trip inside the audible gap. The fetch is not awaited, batches at 100 ids, and its failure is not an error path: `_libraryFallback` still fetches on demand.
- Native mode has no per-item call site to throw at, so an unavailable item mid-queue would strand the queue where one-item mode would have skipped past it. A `mediaPlaybackError` listener drops authority, stops to take it back, and hands the index to `_playAt`. That one is written from the event surface rather than measured: the probes never provoked a failure inside a live native queue. It costs nothing if the event never fires.

### Measurements

Two items over the same track pair, order rotated each pass so queue length and position cannot pose as a difference between designs. MusicKit 3.2526.0-prerelease.x, media-element `currentTime`, 50 ms poll.

| arm | n | median | range | single src |
|---|---|---|---|---|
| rebuild (one item per track) | 11 | 700.0 ms | 451-1000 | 0/11 |
| `items:` catalog + library | 11 | 23.7 ms | 15-37 | 11/11 |
| `items:` 2 catalog | 6 | 350.0 ms | 11-450 | 1/6 |

The invariant that held in every sample is the media-element `src` count rather than any of these medians: one src is always under 40 ms, two are always over 300.

### What is not measured yet

I said on #96 that Phase 1 should fix the segment length it ships and re-measure that shape. This branch does not fix one. It sends the whole queue in a single `setQueue({items:[...]})` (`musickit.html:402`), so the open question is a length sweep rather than a single re-run.

That matters, because length moved the figure on its own in the earlier data: `songs:` gave 26.5 ms at two items and 350.2 ms at three. An all-catalog `items:` queue was also the worst native arm, and I still cannot explain why a library item in second position is the consistent one.

So none of the numbers above should be quoted as this player's behaviour yet. I can run the sweep, the probe is on `proto/96-gapless-probe` and I have a session. I am flagging it here rather than holding the code back, because the sweep may well argue for segmenting the queue, and that is a design question worth settling in review rather than after it.

### Tests

`musickit_test.mjs` covers the native-queue invariants: authority handover, `next`/`prev` delegation and their fallbacks, append against a live native queue, the pre-resolved library copies, and the `mediaPlaybackError` recovery path. 104 assertions, run by pre-commit.

Worth a reviewer's attention: that harness copies logic out of `musickit.html` by hand rather than importing it, so it can stay green while the HTML drifts. `isLibraryID`, `isCatalogID` and `partitionIDs` exist only in the test, inlined from `vibezSetQueue`/`vibezAppendQueue`.

`go test ./...` and `go vet ./...` are clean on linux/amd64.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- **Track transitions no longer rebuild MusicKit's queue**: vibez called `setQueue({songs:[one id]})` at every track boundary, so each transition paid a round trip that landed exactly where the gap is audible. The queue is now built once with `setQueue({items:[...]})` and MusicKit advances it, with `next`/`prev` delegating to `skipToNextItem`/`skipToPreviousItem` and insertions going through `playNext`/`playLater`. `items:` takes resolved descriptors of either kind, so a library song no longer forces the whole queue back to per-track rebuilds. Measured over a two-item queue the boundary gap went from a 700 ms median to 23.7 ms, though the figure moves with queue length and the shipped shape has not been swept yet. Refs #96.
+
 ## [0.7.0] — 2026-08-29
 
 ### Changed
--- a/internal/player/web/musickit.html
+++ b/internal/player/web/musickit.html
@@ -219,13 +219,15 @@
         //
         // There are two modes, and exactly one of them owns advancement:
         //
-        //   native   — one setQueue({songs:[...]}) builds the whole queue and
-        //              MusicKit advances, repeats and shuffles it by itself.
-        //              Navigation is skipToNextItem / skipToPreviousItem; new
-        //              songs go through playLater().
-        //   one-item — setQueue({...}) per track, vibez advances on `completed`.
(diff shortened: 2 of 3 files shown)


simonepelosi:
Thanks @ianswope! Outstanding work on this.

Issue #125, opened 2026-09-03
Can't play explicit titles more than once


ammarasyad:
<img width="2079" height="780" alt="Image" src="https://github.com/user-attachments/assets/c0ff11df-4728-4b3f-aa51-4a417ddf7d6d" />

I was able to play an explicit title **once**, and only if I play the explicit title **first after launching the app**. Afterwards I can't play any explicit titles anymore. Repeat once works though

ianswope:
Thanks for the report. The "only the first one after launch" detail is the useful part.

vibez has no explicit-content handling of its own, so this is coming from MusicKit. Two error codes land in the same place and they mean very different things.

`internal/player/web/musickit.html:608` treats `CONTENT_EQUIVALENT`, `CONTENT_UNAVAILABLE`, `CONTENT_RESTRICTED` and `NOT_FOUND` alike as "this song is unavailable here". It tries your library copy, and if there is not one it skips to the next track without saying anything. That matches your description: the track does not play and no error appears.

Which code you are hitting decides what the fix is:

- `CONTENT_RESTRICTED` is an account setting. Apple refuses explicit content outright and the first-track-only pattern would be coincidence.
- `CONTENT_EQUIVALENT` is MusicKit refusing a queue rebuild mid-session. Every track change re-issues `setQueue`, so the first play of a session is the only one that is not mid-session. That fits exactly, and explicit songs are the likeliest to collide because they have clean counterparts.

Could you reproduce it with the debug log open? Type `:debug-logs` before playing anything. The browser side logs into it, so you should get a line like:

```
[js:doPlay] setQueue CONTENT_EQUIVALENT — trying library
```

The code in that line is what I need.

Either way, skipping silently looks wrong here. A region-locked track is worth passing over quietly; an explicit track the account can play is not.

ammarasyad:
@ianswope Thanks for the reply. I opened debug logs before playing anything

<img width="1380" height="63" alt="Image" src="https://github.com/user-attachments/assets/73d8967c-6299-4541-802b-3c84dc989bfe" />

The line you're looking for isn't logged. So I tried playing an explicit song first thing after opening vibez

<img width="1382" height="458" alt="Image" src="https://github.com/user-attachments/assets/87e463f4-5d3b-4ad6-8876-1b72f175ee63" />

Still nothing like what you want.

ianswope:
[...]

PR #126, opened 2026-09-08
fix(player): deliver every debug log entry, not just the ones that fit

ianswope:
Log entries rode the same subscriber channel as player state: eight slots, sent non-blocking with a bare `default`. Coalescing is right for a position tick and wrong for a log line. A burst around a failed play, which is the case the log exists for, overflowed the buffer and lines disappeared with nothing to show they had.

Each subscriber now owns a queue and a delivery goroutine. State still coalesces to the newest update. Log entries are delivered in order even when the reader is behind, and a backlog past 500 entries reports how many it discarded instead of dropping them quietly. Producers never block.

`State.Log` becomes `State.Logs []string`, since one delivery can now carry several entries. The TUI appends each one.

Surfaced in #125, where two captures of the same code path each showed a different subset of the lines it always emits. This does not fix the skip reported there. It makes the log that would explain it trustworthy.

Tests cover ordering past a full buffer, delivery with no further state behind it, per-subscriber backlog, the cap and its report, and a producer that never blocks on a subscriber that is not reading.


simonepelosi:
Thanks @ianswope! Really nice concurrency design. Separating the delivery semantics between coalescing position ticks and FIFO log lines makes the debug log much more reliable without blocking player producers. Test coverage in broadcast_test.go is very thorough.

Issue #53, opened 2026-05-22
Proposal: prepare AUR packaging for Arch Linux


elijah-rou:
## Proposal

Prepare vibez for distribution through the Arch User Repository, starting with a binary package (`vibez-bin`).

Motivation: I would personally like to install and update vibez through the same Arch/AUR flow I use for the rest of my system, but this is also a useful distribution step for the project. AUR packaging gives Arch users a discoverable install path, integrates updates with their normal package manager, and avoids relying on the curl-based installer as the only Linux-native distribution option. It would also create a good template for other repository formats later, for example Homebrew, Nixpkgs, Fedora COPR, openSUSE OBS, Debian/Ubuntu PPA, or native distro packages.

## Suggested first package

I suggest starting with `vibez-bin`, rather than a source-built `vibez` package.

Release builds currently embed the Apple Music developer token and Last.fm app keys via ldflags/garble. A source AUR build cannot reproduce that without requiring each user to provide their own Apple Developer credentials, so it would not be equivalent to the official release binary.

## Proposed packaging shape

Create an AUR package that installs the official Linux amd64 release artifact:

- package name: `vibez-bin`
- `provides=('vibez')`
- `conflicts=('vibez')`
- `arch=('x86_64')`
- `license=('MIT')`
- install binary to `/usr/bin/vibez`
- install desktop file to `/usr/share/applications/io.github.simonepelosi.vibez.desktop`
- install icon to `/usr/share/icons/hicolor/512x512/apps/io.github.simonepelosi.vibez.png`
- install license to `/usr/share/licenses/vibez-bin/LICENSE`

Likely runtime dependencies:

- `webkit2gtk-4.1`
- `gtk3`
- `gstreamer`
- `gst-plugins-base`
- `dbus`
- `xdg-utils`
- `tar`

Potential optional dependencies:

- `gst-plugins-good`: additional GStreamer playback support for the WebKit/GStreamer fallback
- `gst-plugins-bad`: additional GStreamer playback support for the WebKit/GStreamer fallback
- `gst-libav`: codec support for the WebKit/GStreamer fallback
- `google-chrome`: possible future system Chrome backend instead of downloading Chrome at first run

## Repo/release changes that would make this cleaner

1. Include packaging assets in the GitHub release tarball.
   - Current GoReleaser archive appears to include only `vibez`.
   - Including `LICENSE`, the desktop file, and icon would let the AUR `PKGBUILD` install from a single release artifact.

2. Align desktop entry naming.
[...]

PR #127, opened 2026-09-10
build(release): ship the desktop file and icon in the release archive

ianswope:
`archives.files` now includes `flatpak/io.github.simonepelosi.vibez.desktop` and `assets/logo.png`, so a distro package can install the binary, desktop entry, and icon from the one release tarball instead of fetching two files from the tag separately. They are the same two files `scripts/install.sh` downloads and the Flatpak manifest installs.

Setting `files` replaces GoReleaser's default list, so `LICENSE*`, `README*` and `CHANGELOG*` are restated; the archive contents are otherwise unchanged. The darwin archive picks up the two files as well, since both builds share the one archive block. A second block to keep 270 KB out of the macOS tarball did not seem worth it, but it is a small change if you would rather.

`goreleaser check` passes. It reports the pre-existing `builds.gobinary` deprecation, which is unrelated.

Refs #53, item 1.


simonepelosi:
Thanks @ianswope! Bundling the desktop entry and logo directly in the release tarball makes packaging for distro maintainers much cleaner and avoids separate raw fetches. Agreed that keeping a single `archives` block rather than splitting macOS out is the right tradeoff.

Just needs a quick rebase on `main` to catch up with #131. LGTM!

PR #128, opened 2026-09-10
fix(assets): install the desktop entry under the id MPRIS advertises

ianswope:
MPRIS advertises `DesktopEntry` as `io.github.simonepelosi.vibez`, and the spec resolves that to `io.github.simonepelosi.vibez.desktop`. `InstallDesktopEntry` wrote `vibez.desktop`, so a desktop that found the player over MPRIS could not find its entry, which is the one thing the function exists for. The install script and the Flatpak manifest use the right name, which masked this for anyone who installed that way; `make install` and source builds were affected.

Changes:

- The id is one constant, `assets.AppID`, used by the embedded desktop file, both installers, and the MPRIS property.
- The desktop entry and icon are written as `io.github.simonepelosi.vibez.desktop` and `io.github.simonepelosi.vibez.svg`, and the embedded entry's `Icon=` matches. The `vibez.desktop` and `vibez.svg` earlier versions left in `~/.local/share` are removed.
- The self-install is now a fallback. If an entry or icon with that name already exists in `$XDG_DATA_HOME` or any `$XDG_DATA_DIRS` entry, nothing is written, so a distro package, a Flatpak export, or the install script's copy is left alone. This matters for the install script in particular: with the names aligned, writing the embedded `NoDisplay=true` entry on every launch would have replaced its visible one. The fallback is written once and found by the same check afterwards.
- `$XDG_DATA_HOME` is honoured for the write, defaulting to `~/.local/share` as before.

Tests cover the fallback write, the skip for a system entry, an existing user entry being left untouched, legacy cleanup, and the icon lookup preferring an installed PNG.

Refs #53, items 2 and 3.


simonepelosi:
Thanks @ianswope! Excellent catch on the MPRIS name alignment and especially the fallback guard to avoid clobbering the visible entry installed by `install.sh`.

Implementation and test coverage are very clean. Only note is that it needs a quick rebase on `main` to catch up with #131 (merges cleanly with no conflicts).

LGTM, happy to merge once rebased.

PR #129, opened 2026-09-10
fix(cdp): honour VIBEZ_CHROME_PATH on amd64

ianswope:
`findSystemBrowser` has honoured `VIBEZ_CHROME_PATH` and `CHROME_PATH` since #94, but only arm64 ever called it. On amd64 `ChromePath` returned the privately extracted Chrome unconditionally, so an x86_64 user could not point vibez at a browser they already have, and a packaged install would download about 130 MB on first run regardless.

The arm64 gates now ask `useSystemBrowser()`, which is true on arm64 as before and on any arch when one of the two variables is set. Everything the arm64 path already does follows: no download, no `vibez-helper` hard link (it cannot link into a system location), the persistent profile so a component-updated Widevine CDM is picked up, and the system CDM lookup. `widevineCDMDir` also checks `_platform_specific/linux_x64`, the name Chromium uses on amd64. An unusable override fails in `EnsureBrowser` with the variable named, before anything is downloaded. Without the variables set, amd64 behaves exactly as it did.

This does not make the system browser the default on amd64. The private Chrome is what makes Widevine work with nothing installed, and a distro Chromium often has no CDM.

The arm64 test helpers move to `browser_system_test.go` so the new amd64 tests can share them. README gets a sentence on the variable.

Refs #53, item 4.


simonepelosi:
Thanks @ianswope, clean refactor and solid test coverage. The override logic works as expected without regressing the default amd64 flow.

A few quick observations:

1. **Error message on amd64**: If an override browser lacks Widevine CDM (or warmup times out), `ensureSystemBrowser` outputs `chromeInstallHelpARM64`, which references Arch Linux ARM and suggests setting `VIBEZ_CHROME_PATH` (which was already set). On amd64, it would be much clearer to drop the ARM-specific reference and mention they can unset the variable to let vibez use its bundled Chrome with Widevine.
2. **Generic `CHROME_PATH`**: `CHROME_PATH` is often exported in dev shells for Puppeteer/Karma pointing to a distro Chromium without Widevine. On amd64, those users will now automatically hit the system browser path and fail on missing Widevine. Just worth keeping in mind if anyone reports unexpected CDM errors.
3. **Doc comment mismatch in `widevineCDMDir()`**: The comment says callers can launch without `--widevine-path` when it returns `""`, but `ensureSystemBrowser` actually treats `widevineCDMDir() == ""` as fatal. A small nit to align the doc with the check.
4. **Rebase**: Needs a quick rebase on `main` to pull in #131 (clean merge, no conflicts).

LGTM overall, happy to merge once rebased and after checking the error message.

ianswope:
Rebased on `main`, and both code points are in as a follow-up commit.

**1. Error message on amd64.** `chromeInstallHelpARM64` is now `systemBrowserHelp()`, covering the three sites that used it: `findSystemBrowser`, the CDM check in `ensureSystemBrowser`, and the warm-up timeout. With an override set on amd64 it says to point the variable at a browser with a Widevine CDM, or unset it to use the Chrome vibez downloads, which bundles one. The Arch Linux ARM reference is gone from that case. On arm64 unsetting means falling back to system discovery rather than a bundled Chrome, so that case says that instead of promising something that does not exist there. No override keeps the original wording.

**3. `widevineCDMDir()` doc comment.** Corrected, and moved: it had drifted onto `var widevineSystemDirs`, so it was not attached to the function it describes. Both claims are true, of different callers, and it now says which is which. `chromeLaunchArgs` omits `--widevine-path` on `""` and lets Chromium self-locate; `ensureSystemBrowser` treats `""` as fatal before it ever gets there.

Three tests, each of which fails against the old constant: the amd64 override advice, the no-override advice, and an `EnsureBrowser` case asserting that a missing CDM blames the override rather than printing ARM guidance.

**2. Generic `CHROME_PATH`.** Left alone deliberately, but I agree it is a real trap, and it is new on amd64 Linux specifically, where the bundled Chrome was previously unconditional. Narrowing the amd64 override to `VIBEZ_CHROME_PATH` only would fix it, at the cost of the `CHROME_PATH` parity `browser_darwin.go` already has (it checks `VIBEZ_CHROME_PATH` then `CHROME_PATH`, `browser_darwin.go:33-38`). That is a behaviour call rather than a cleanup, so I would rather you picked. Say which and I will follow up separately.

ianswope:
Bumping point 2 from your review, the one thing here that never got decided. It has shipped in v0.8.0 and v0.9.0 since, so the dev-shell case is live rather than hypothetical.

Where it stands: `browserOverride` checks `VIBEZ_CHROME_PATH` then `CHROME_PATH` (`internal/player/cdp/browser.go:50`), and on amd64 either one now diverts vibez off the bundled Chrome. Someone with `CHROME_PATH` exported for Puppeteer or Karma gets a distro Chromium with no CDM, and playback fails for a user who never set out to configure vibez at all. It surfaces only as a Widevine error, which does not point at the cause.

The two options, unchanged:

- **Narrow the amd64 override to `VIBEZ_CHROME_PATH`.** Removes the trap. Costs the parity with `browser_darwin.go:32-38`, which honours both, and leaves the two Linux arches reading different variables.
- **Leave it.** Every platform reads the same two names, and the trap stays.

If you would rather not spend a decision on it, I would narrow it. `CHROME_PATH` is a variable other tools set about themselves; `VIBEZ_CHROME_PATH` is one a user sets about vibez. The parity argument is real, but darwin never had a bundled browser to be diverted off: `findChromePath` has always required a user-installed Google Chrome (`browser_darwin.go:39-52`), so `CHROME_PATH` there picks between browsers the user chose to install. On amd64 it overrides the one vibez guarantees, which is why this is new rather than longstanding.

Still your call. Either way it is a small follow-up and I will send it.

PR #132, opened 2026-09-13
fix(apple): send catalog search to the supported host

### LINKED ISSUE #93: [Bug] Apple Music discovery mode fails with "[vibe] search error: no results"

ianswope:
Closes #93.

Catalog song search went to `amp-api.music.apple.com`, picked because it is the only endpoint returning `extendedAssetUrls` and so the only one whose results could be filtered for streamability. It has been rejecting developer/user token pairs that the supported endpoints accept, last re-verified on 3 August.

#118 added a retry against `api.music.apple.com`, which stopped this being fatal, but the rejection still has to happen on every search before the retry does. A failing songs leg is not fatal to `Search` (`apple.go:479` errors only when all three legs fail), so before #118 discovery was refilling from the user's library alone: unrelated tracks for a user who happened to have the artist, an empty queue for one who did not.

This makes the supported host the only path, as you suggested on the issue:

- catalog search moves to `api.music.apple.com`
- `extend=extendedAssetUrls` is dropped
- the filter becomes `PlayParams != nil && Kind == "song"`

That endpoint cannot report streamability, so playback becomes the final authority on storefront availability. Catalog album and playlist tracks have been queued on exactly those terms since #82 moved them to the same host, so this makes search consistent with them rather than loosening anything new. The measurements on the issue from 3 August are the basis: 25 songs per artist from the supported host, all with `playParams.kind == "song"`, none with a non-empty `extendedAssetUrls.plus`.

**Removed:** `extendedAssetURLs` and `hasStream()`, which existed only for the filter that is gone.

**Left in place, deliberately:** `newCatalogRequest`, `catalogBaseURL` and `defaultCatalogURL`. Nothing reaches them now, since every `fetchSongsPaginated` caller passes `isCatalog=false` once #82 and this change are both in. I added a comment saying so rather than deleting them, because whether vibez is finished with amp-api is your call and not a consequence of this fix. Happy to strip it in a follow-up.

**Tests.** Three fail against the current implementation and pass here: the endpoint assertion (no `extend=`, exactly one songs request), a purchase-only track surviving the filter, and the playability filter reduced to `playParams` and `kind`. A fourth covers a rejected songs leg surfacing as a warning rather than an empty result; it passes either way, and is there because there is no retry left to mask it.


simonepelosi:
Thanks @ianswope! Excellent execution on this. Migrating catalog song search directly to `api.music.apple.com` and aligning the playability filter with `PlayParams != nil && Kind == "song"` makes it consistent with album/playlist tracks and cleanly unblocks discovery mode.

Also agree with keeping `newCatalogRequest` and the base URL constants in place with explanatory comments rather than stripping them in this fix.

Test coverage is thorough and all CI checks pass. LGTM!

PR #133, opened 2026-09-14
fix(mpris): export our own Properties handler instead of godbus's

### LINKED ISSUE #110: MPRIS: PropertiesChanged emits godbus's mutable stored value, racing with the next update

ianswope:
Closes #110.

As worked out on the issue, `Emit: prop.EmitFalse` would only have closed one of two legs. `prop.Export` keeps one allocation per property and merges each update into it with `dbus.Store`; for `a{sv}` that is an in-place `storeMapIntoMap`. It hands that live map to the encoder from `emitChange`, and also from `Get`/`GetAll`, which return `reflect.ValueOf(prop.Value).Elem().Interface()` and release `p.mut` before the reply is encoded in the output goroutine. The second leg involves no `PropertiesChanged`, so no `Emit` setting reaches it.

Both interfaces now go through `properties` in this package, root included as you asked:

- `Get`/`GetAll` build a fresh outer map per call
- Metadata is replaced wholesale, never merged, so it is immutable once it escapes
- `flush` emits one batched `PropertiesChanged` instead of the 6 to 8 per track change
- `Position` is no longer in that signal. The spec marks it `EmitsChangedSignal=false` and announces jumps with `Seeked`, which `flush` already sends, so the previous `EmitInvalidates` was wrong; the comment in `flush` already described the new behaviour.

`trySetProperty` and its `recover` go with `SetMust`, the only thing that panicked. A failed emit still calls `markUnavailable`.

**Test.** `TestServer_GetAllDuringFlushIsRaceFree` drives real `GetAll` calls over the session bus against 60 track changes. Against `main`'s `server.go` it fails three runs out of three with `storeMapIntoMap` against `handleCall`; on this branch `-race -count=10` is clean across the package.

Worth noting for #88: a concurrent map read and write is a runtime throw, not a panic, so this was a candidate for the exit-2 closures there. zettzu's A/B with MPRIS disabled ruled it out as their cause, so this does not close that one.


simonepelosi:
Thanks @ianswope! Outstanding work on this. Replacing `godbus/prop.Export` with our own `org.freedesktop.DBus.Properties` handler cleanly resolves the `storeMapIntoMap` race condition, and invoking `onSet` outside the lock prevents any deadlock risks with the player.

Also a big plus on the single batched `PropertiesChanged` signal in `flush()`, properly leaving out `Position` per spec's `EmitsChangedSignal=false`, and migrating both root and player interfaces for consistency. Test coverage with the race detector is great.

LGTM!

PR #135, opened 2026-09-14
fix(cdp): retain driver output and attach it to driver failures

ianswope:
Picks up @elijah-rou's #103, which they offered to hand over. The design is theirs and they are the co-author on the commit; #103 can be closed in favour of this.

It is not a rebase of that branch. `driverio.go` was rewritten by #99 after #103 was written, so `driverOutput()` no longer exists and their diff no longer applies. This re-applies their design to the current shape and keeps the parts of both that turned out to be right.

### What was wrong

`driverStderr()` returned `io.Discard`. Handing node a pipe rather than an `*os.File` is required: `os/exec` passes a file descriptor straight through to the child, and the driver's exit handler restores a termios snapshot taken before BubbleTea switched the terminal to raw mode, which is what leaves a shell with no echo and no ctrl+c after vibez quits. Discarding the output was the cheapest way to get a pipe, and it threw away the only account of why a driver failed to start.

### What changed

`boundedDriverOutput` retains the last 64 KiB and is still a plain `io.Writer`, so the child keeps its pipe. `addDriverOutput` appends that tail to a failure and leaves the error alone when the driver said nothing, so a silent failure does not gain an empty section.

Every Playwright entry point in the package now builds its options through `newDriverRunOptions`. That was elijah-rou's point on #103 and it is the part worth keeping: a missed call site becomes a missing constructor call rather than a silently absent field. `TestDriverRunOptions_IsTheOnlySourceOfRunOptions` enforces it by scanning the package's non-test sources for a bare `playwright.RunOptions{`, so the next split of `EnsureBrowser` cannot quietly reintroduce one.

The logger is one shared instance rather than one per call. playwright-go keeps it in package state (`run.go:36`, reassigned at `run.go:455`) and its default writes to stderr, which it uses for navigation errors, WebSocket failures and BindingCall rejections at times vibez cannot predict.

### Two notes on scope

**The install path gains nothing visible today.** I traced it rather than assuming: `d.log` is gated on `Verbose`, which is false here, and `SkipInstallBrowsers: true` skips `installBrowsers`, the only other writer. So `playwright.Install` currently produces no output to capture. It goes through the constructor for consistency and so that a future `Verbose` or a non-skipped browser install cannot reach the terminal. The behaviour change is on the long-lived driver's stderr, which playwright-go wires to `cmd.Stderr` at `transport.go:113`.

**Both comments on #103 were right, about different lines.** The doc comment on `driverLogger()` said the concern was `log.SetOutput` in `transformRunOptions` (`run.go:452`); elijah-rou said playwright-go stores the logger in package state (`run.go:455`). Both are true and both matter, so the new comment says so and cites each.

### Testing

10 tests in `driverio_test.go`. The two existing ones are kept, with the terminal check retargeted at the constructor and the standard-logger check still exercising the real dependency rather than trusting a read. New: the shared logger instance, three `boundedDriverOutput` cases (tail preserved across two writes that only overflow together, a single oversized write, and `Write` reporting the full length it was given so callers never see a short count), three `addDriverOutput` cases (nil error, silent driver, and `errors.Is` still reaching the cause through the wrap), and the literal scan.

I checked two of them fail when the invariant breaks rather than trusting that they pass: nilling `Stdout` fails the terminal test, and adding a bare `playwright.RunOptions{}` to `driver.go` fails the scan, each with the intended message.

Clean on both platforms, since the diff touches `browser_darwin.go` as well:

- linux/amd64: `go test ./...`, `go vet ./...`, `gofmt -l .`, `golangci-lint 2.11.4` (0 issues)
- darwin/arm64, go1.26.1, `CGO_ENABLED=1`: `go vet ./...` clean, `go test ./...` 18 packages ok, `gofmt` clean


simonepelosi:
Thanks @ianswope and @elijah-rou! Excellent pickup and clean adaptation to the current post-#99 codebase.

Key highlights:
- Retaining the bounded 64 KiB buffer via `boundedDriverOutput` provides the diagnostics we need on startup crashes while strictly keeping node on a pipe rather than an `*os.File`, preserving termios/TTY integrity.
- Centralizing options in `newDriverRunOptions` and enforcing it with `TestDriverRunOptions_IsTheOnlySourceOfRunOptions` is a great guard against future regressions.
- The shared `playwrightLogger` keeps the process-wide standard logger intact.

Test coverage is thorough and all CI checks pass. LGTM!

simonepelosi:
I'm merging this and closing the #103

PR #140, opened 2026-09-15
feat: browserless Apple Music playback on macOS via native Widevine CDM & AVFoundation

simonepelosi:
## Overview

This PR builds directly on top of PR #138 (`exp/browserless-player`) as a prerequisite, bringing the high-efficiency browserless Apple Music playback engine to **macOS (Darwin)** and adding an **automatic Widevine CDM downloader mirroring Firefox**.

### Key Changes
1. **Universal Audio Sink Abstraction**:
   - Refactored `internal/player/browserless/player.go` to depend on a lightweight, platform-agnostic `audioSink` interface rather than concrete `gst.Player`.
   - Preserved `internal/player/gst` on Linux (`sink_linux.go`).
   - Implemented native macOS audio output using Apple's `AVFoundation` (`AVPlayer`) framework (`sink_darwin.go`, `sink_darwin.m`, `sink_darwin.h`) with zero external package dependencies (`brew install gstreamer` not needed).
   - Provided no-cgo stubs (`sink_darwin_nocgo.go`, `cdm_nocgo.go`) so CI and test runs with `CGO_ENABLED=0` pass cleanly.

2. **Automated Widevine CDM Downloader (Mirroring Firefox)**:
   - If no supported browser is found on the system, Vibez automatically queries the official Google component CDN (via Mozilla's `widevinecdm.json` manifest with hardcoded Google CDN fallbacks) and downloads the official Google `.crx3` package for the current OS/architecture:
     - macOS Apple Silicon (`mac_arm64`)
     - macOS Intel (`mac_x64`)
     - Linux x86_64 (`linux_x64`)
   - Unpacks and extracts `libwidevinecdm.dylib` / `libwidevinecdm.so` in ~2 seconds into the user cache directory with decompression bounds.
   - Completely eliminates any requirement to have Google Chrome, Brave, or any browser installed on the host system.

3. **Expanded Local CDM Auto-Discovery**:
   - Discovers existing `libwidevinecdm` installations across:
     - Google Chrome (`mac_arm64` & `mac_x64`)
     - Brave Browser (`mac_arm64` & `mac_x64`)
     - Microsoft Edge (`mac_arm64` & `mac_x64`)
     - Arc Browser (`mac_arm64` & `mac_x64`)
     - Mozilla Firefox profiles on both macOS and Linux (`~/Library/Application Support/Firefox/Profiles/*/gmp-widevinecdm/*/` and `~/.mozilla/firefox/*/gmp-widevinecdm/*/`)
     - User cache & config directories
   - Added Darwin CGo flags in `cdm.go`: `#cgo darwin LDFLAGS: -ldl -lc++`.

4. **Platform Dispatch**:
   - Updated `cmd/root_darwin.go` to use `browserless.Available()` and route to `runBrowserlessFlow`, with TUI notification when downloading the CDM component on initial launch.
   - Factored out platform-specific media keys (MPRIS on Linux) and backend status bar labels.

### Prerequisites
- **Prerequisite PR**: #138 (`exp/browserless-player`)


ianswope:
This does not build on macOS. I think the reason nobody caught it is that
ci.yml triggers on `pull_request: branches: ["main"]`, and this PR targets
exp/browserless-player, so no checks have ever run against it. Worth adding the
base branch to that trigger.

Tested on macOS 27.0, Go 1.27.1 darwin/arm64, Apple clang 21, at dca2f21.

1. `CGO_ENABLED=0 go build ./...`, which is exactly what the macOS CI job runs,
fails in `internal/player/browserless`:

    streamer.go:390: undefined: cdm.Subsample
    streamer.go:397: s.cdmEngine.DecryptSubsamples undefined

`cdm_nocgo.go` was written against the CDM API as it stood before the #138
review fixes, and merging #138 in did not bring it forward. The no-cgo stubs
need `Subsample` and `DecryptSubsamples` too.

2. With cgo enabled the build fails in the preamble. `sink_darwin.go` declares

    extern void vibezDarwinOnError(void* userData, const char* errMsg);

while cgo generates `char*` from the Go signature, and clang 21 makes the
conflict an error rather than a warning. That may be why it built for you.
What builds here: declare the extern with `char*` and cast at the call site,
`(vibez_err_cb_t)vibezDarwinOnError` in `vibez_create_darwin_sink`.

3. Past that, `cmd/root_darwin.go:20` calls `cdp.Available`, which is defined in
`internal/player/cdp/browser.go` under `//go:build linux`. That one is
unconditional; no toolchain will accept it.

4. `go vet` fails on `sink_darwin.go:41`: `possible misuse of unsafe.Pointer`.
A `cgo.Handle` is an integer, not a pointer, so `unsafe.Pointer(uintptr(h))` is
not a legal conversion. Pass it as `C.uintptr_t` and take `uintptr_t` on the C
side. It happens to work today, but the macOS CI vet step runs with cgo on, so
it will fail there once CI does run.

5. `TestDownloadCDMReal` hardcodes `libwidevinecdm.so` and is tagged
`linux || darwin`, so it fails on every macOS run, since the mac CRX carries a
`.dylib`. Separately it is a live network test behind nothing but `-short`, so
a plain `go test ./...` pulls ~14 MB from Google's CDN and writes out an 18.7 MB
proprietary library. I would not have CI doing that on every push.

The one I actually want to talk about:

6. The downloader verifies nothing it downloads. Mozilla's manifest carries
`hashValue` and `filesize` for every platform, with `hashFunction: sha512` at
the top, and `resolveDownloadURL` reads only `fileUrl`. `extractLibraryFromCRX`
then skips the CRX3 signed header by its length without checking the signature
or the publisher key. So the chain is: a JSON file fetched from a GitHub mirror
names a URL, whatever is at that URL gets unpacked, and the result is dlopen'ed
as native code in our process. Verifying the sha512 that is already in the
response is a few lines, and it is the half of "mirroring Firefox" that
actually matters, because verifying is the thing Firefox does.

7. Related, on consent. On darwin `Available()` is true whenever the platform
has a download URL at all, so `root_darwin.go:31` sends any Mac without Chrome
into the browserless engine and fetches Widevine on first launch with no opt-in.
The Linux path in #138 is more conservative and requires a CDM already on disk.
Downloading a proprietary binary onto someone's machine deserves an explicit
yes, and it is also what decides how this lands in #53: a distro package cannot
ship that blob, and probably should not fetch it silently either.

Minor: the 64 MiB `io.LimitReader` truncates rather than erroring, so an
oversized library would be written short and fail at dlopen with nothing
pointing at why. And `mirrorUrls` in the manifest go unused.

PR #141, opened 2026-09-16
fix(local): say why a local library came back empty

### LINKED ISSUE #134: Local mode gives no sign when a file's format is unsupported on that platform
`--local` skips any file whose extension is not in the platform's `supportedExts` map and says nothing about it (`internal/provider/local/provider.go:51`):

```go
ext := strings.ToLower(filepath.Ext(path))
if !supportedExts[ext] {
    return nil
}
```

The map differs by platform. Linux has `.ogg`, darwin does not (`exts_linux.go`, `exts_darwin.go`), because the CoreAudio backend does not decode it and GStreamer does. That part is right.

What is missing is any feedback. `scan()` counts nothing, `GetLibraryTracks` returns whatever it found, and `cmd/root.go:124` launches the TUI on an empty slice without comment. A macOS user whose library is OGG gets a TUI that says `Local mode · playing from /home/user/Music` and lists nothing.

That is indistinguishable from the other ways to reach an empty library:

- the path is wrong
- the directory really is empty
- the files failed `os.Open` in `trackFromFile`, which also returns `nil` and drops them (`provider.go:56`)

So the one case that has a clear explanation and a clear remedy looks exactly like the cases that have neither.

It is harder to work out than it sounds, because the platform split is not written down anywhere the user will look. The README does not mention `--local` or `--music-dir` at all, `--help` says only "run with local music files (no Apple account required)", and #61's title is "local tracks provider (M4A, FLAC, MP3, OGG)" with no platform qualifier. The v0.9.0 release notes are currently the only place stating that OGG is Linux only.

Smallest useful fix: have `scan()` keep a count of files skipped for an unsupported extension and surface it. A debug log line costs nothing. Something visible when the resulting library is empty would do more, along the lines of:

```
no playable tracks in /home/user/Music (skipped 412 files, .ogg is not supported on macOS)
```

Happy to send it if the shape looks right. Not urgent and not a regression: it arrived with #62 rather than breaking anything that worked.

ianswope:
Closes #134.

### What was wrong

`scan()` returns `nil` for any file whose extension is missing from the platform's `supportedExts` map (`provider.go:51`) and counts nothing. `GetLibraryTracks` returns what it found, `cmd/root.go` launches the TUI on an empty slice, and nothing anywhere says a file was dropped.

The map differs by platform on purpose: `.ogg` plays on Linux through GStreamer and not on macOS through CoreAudio. So a macOS user whose library is OGG gets a TUI headed `Local mode · playing from ~/Music` listing nothing, which is exactly what a mistyped path, an empty directory, and files that will not open all produce as well. Only the format case has a remedy, and it is the one case that looks like the others.

The supported set is also not written down anywhere a user would look: the README does not mention `--local` or `--music-dir`, `--help` says only "run with local music files (no Apple account required)", and #61's title reads "M4A, FLAC, MP3, OGG" with no platform qualifier. The v0.9.0 release notes are the only place that states OGG is Linux only.

### What changed

`scan()` keeps two counts: files skipped for an unsupported extension, keyed by extension, and files with a supported extension that would not open. `ScanNotice()` turns them into one line.

```
no playable tracks in ~/Music: macOS plays .flac, .m4a, .mp3; skipped 412 files (410 .ogg, 2 .jpg)
no playable tracks in ~/Music: the directory is empty
no playable tracks in ~/Music: 3 files could not be read
indexed 12 of 14 files in ~/Music: Linux plays .flac, .m4a, .mp3, .ogg; skipped 2 files (1 .cue, 1 .jpg)
```

It returns `""` when every file was indexed, so the ordinary case stays silent. `cmd/root.go` passes it as `Options.ScanNotice`; the model logs it on every local launch and puts it in the status bar only when the scan produced no tracks at all.

Four details that are deliberate rather than incidental:

- **The playable set comes before the skipped detail, and `$HOME` is elided to `~`.** The status bar truncates from the right, so at 80 columns what survives is `no playable tracks in ~/Music: Linux plays .flac, .m4a, .…`, which is the half the user can act on. The path is the part they already know.
- **The extension list is capped at three, commonest first, with `+N more`.** A music directory routinely holds artwork, cue sheets and logs beside the audio, so the tail is long and is never the answer.
- **`platformName` is a new const beside `supportedExts`** in `exts_linux.go` and `exts_darwin.go`, so the name in the message and the set it describes cannot drift apart.
- **Files that would not open are counted separately and get no "plays" clause.** Nothing about the format explains them, so the message does not blame it.

### One note on scope

`nowPlayingTextLines` carried its own copy of `statusLine`'s formatting, without the truncation (`model.go:3275`). That is the path the empty-library case always renders through, since nothing is playing, so a message long enough to explain itself would have overflowed the box instead of being cut to the width. It now calls `statusLine`, which is the same rendering plus the truncation, and deletes the duplicate.

That is adjacent to #134 rather than part of it. I would rather not ship the notice without it, but say the word and I will split it into its own PR.

### Testing

11 new tests. Nine in `internal/provider/local/notice_test.go`: the four message shapes, the empty directory, a file with no extension, the three-extension cap, a file that will not open, and the `~` elision. Two in `internal/tui/model_test.go`: shown in the status bar when the library came back empty, logged only when it did not.

Two of them are built to survive the environment rather than assume it. The unreadable-file case uses a dangling symlink instead of `chmod 000`, so it does not depend on the test running as a non-root user, which is not true in a container. The OGG case asserts against a per-platform const, so one test expects silence on Linux and the full notice on macOS.

- linux/amd64: `go test ./...`, `go vet ./...`, `gofmt -l .`, golangci-lint 2.11.4, 0 issues.
- darwin/arm64, go1.27.1, macOS 27.0: `go test ./...` clean with cgo on, `go vet ./...` clean, and `CGO_ENABLED=0 go build ./...` clean, which is what the macOS CI job runs.
- Ran it for real on Linux: `--local` against a directory of nine `.wma` plus a jpg and a cue, at 100 and 80 columns. The notice appears, the box is intact at both widths. Against twelve `.ogg` plus the same two, twelve tracks queue and the notice appears in `:debug-logs` only.


simonepelosi:
Thank you for this thorough and thoughtful PR! The UX details here are fantastic — especially placing the actionable platform capability at the start of the message for status bar truncation, capping the extension list, and cleaning up the rendering in `nowPlayingTextLines`. The test coverage across both platforms is also top-notch. Really great work!

PR #142, opened 2026-09-16
fix(ci): run checks on pull requests that do not target main

ianswope:
`ci.yml` triggers on `pull_request: branches: ["main"]`, so a pull request targeting anything else never runs a check. #140 is the case that surfaced it: open since 09-15 against `exp/browserless-player`, and not one check has run on it, on either platform.

This drops the filter rather than adding the branch by name. The next stacked PR will have a different base, and the failure mode is silent: no check appears, nothing goes red, and the PR looks exactly like one that passed.

`push` keeps its filter, so this does not double up on pushes to main.

One note on when it takes effect. A `pull_request` run uses the workflow file from the merge of head into base, so an open PR only picks this up once its base branch carries it. #140 will need `exp/browserless-player` refreshed from `main` before checks appear on it.


simonepelosi:
Thanks for catching this! Makes total sense to drop the branch filter on `pull_request` so stacked PRs against feature branches get full CI coverage. Clean fix, thanks for the great work!

